PR #33, opened 2016-03-31
Warns if the written log files exceed a certain limit

andyHa:


src/main/java/sirius/kernel/Setup.java
```diff
@@ -330,6 +326,28 @@ public void cleanOldLogFiles(long retentionInMillis) {
     }
 
     /**
+     * Estimazes the total size of all log files.
```
sbi-scireum: Estimates

PR #34, opened 2016-04-01
Make urlEncode not fail when a null String is entered

sbi-scireum:
Make urlEncode not fail when a null String is entered

### DIFF
--- a/src/main/java/sirius/kernel/commons/Strings.java
+++ b/src/main/java/sirius/kernel/commons/Strings.java
@@ -170,14 +170,15 @@ public class Strings {
      * @param value the value to be encoded.
      * @return an url encoded representation of value, using UTF-8 as character encoding.
      */
-    public static String urlEncode(String value) {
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/kernel/commons/Strings.java
```diff
@@ -170,14 +170,15 @@ public static String firstFilled(String... values) {
...
-        } catch (UnsupportedEncodingException e) {
-            // Cannot happen if Java-Version is > 1.4....
-
-            return value;
+    public static String urlEncode(@Nullable String value) {
+        if (isFilled(value)) {
+            try {
+                return URLEncoder.encode(value, Charsets.UTF_8.name());
+            } catch (UnsupportedEncodingException e) {
+                // Cannot happen if Java-Version is > 1.4....
```
andyHa: Exceptions.ignore(e) ...

PR #52, opened 2017-01-18
Adds min and max functions to amount which handle null and NOTHING gracefully

andyHa:
Adds min and max functions to amount which handle null and NOTHING gracefully

### DIFF
--- a/src/main/java/sirius/kernel/commons/Amount.java
+++ b/src/main/java/sirius/kernel/commons/Amount.java
@@ -515,7 +515,7 @@ public class Amount implements Comparable<Amount> {
      * If either of the amounts is empty, the filled one is returned. If both are empty, an empty amount is returned.
      *
      * @param other the amount to compare against
-     * @return the amount with the lower value or an empty amount, of both amounts are empty
+     * @return the amount with the lower value or an empty amount, if both amounts are empty
      */
     @Nonnull
     public Amount min(@Nullable Amount other) {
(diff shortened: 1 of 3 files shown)


src/main/java/sirius/kernel/commons/Amount.java
```diff
@@ -509,6 +509,52 @@ public int hashCode() {
...
+        return this.amount.compareTo(other.amount) < 0 ? this : other;
+    }
+
+    /**
+     * Compares this amount against the given amount and returns the one with the higher value.
+     * <p>
+     * If either of the amounts is empty, the filled one is returned. If both are empty, an empty amount is returned.
+     *
+     * @param other the amount to compare against
+     * @return the amount with the higher value or an empty amount, of both amounts are empty
```
meg1502: 'of both' = 'if both'

src/main/java/sirius/kernel/commons/Amount.java
```diff
@@ -509,6 +509,52 @@ public int hashCode() {
         return amount != null ? amount.hashCode() : 0;
     }
 
+    /**
+     * Compares this amount against the given amount and returns the one with the lower value.
+     * <p>
+     * If either of the amounts is empty, the filled one is returned. If both are empty, an empty amount is returned.
+     *
+     * @param other the amount to compare against
+     * @return the amount with the lower value or an empty amount, of both amounts are empty
```
meg1502: 'of both' = 'if both'

PR #56, opened 2017-02-10
aha/BetterComposites

andyHa:


src/main/java/sirius/kernel/di/transformers/Composable.java
```diff
@@ -21,19 +21,40 @@
...
  * Next to the <tt>Adapter Pattern</tt> supported via {@link Transformers}, this implementation permits to
  * add components via the <tt>attach</tt> methods which can later be queried via <tt>tryAs, as</tt> and <tt>is</tt>.
+ * <p>
+ * This class can be used as base class or embedded into another class (using {@link #Composable(Object)}) to
+ * make it <tt>Transformable</tt>.
+ * </p>
  */
 public class Composable implements Transformable {
 
+    private Object target;
```
Tobsucht: should be named "source"?

PR #70, opened 2017-04-10
Makes it possible to add ellipsis dots when limiting Strings

D3nnisH:


src/main/java/sirius/kernel/commons/Strings.java
```diff
@@ -275,11 +275,24 @@ public static String urlEncode(@Nullable String value) {
...
+     * @return a part of the string representation of the given <tt>input</tt>. If input is shorter
+     * than <tt>length</tt>, the full value is returned. If input is <tt>null</tt>, "" is returned.
+     */
+    public static String limit(@Nullable Object input, int length, boolean showEllipsis) {
         if (isEmpty(input)) {
             return "";
         }
         String str = String.valueOf(input).trim();
-        return str.substring(0, Math.min(length, str.length()));
+        return str.substring(0, Math.min(length, str.length())) + (showEllipsis && str.length() > length ? "…" : "");
```
andyHa: brainoverload :-P shouldn't we go to the boolean store and buy an if?

D3nnisH: then we needed to extract the substring part into a local variable (aka store it in `str`) before the if to avoid duplicate code. But then we cannot check `str.length() > length` anymore hence we needed an extra local boolean variable :/

PR #82, opened 2017-05-08
Improves the Trie class

D3nnisH:
adds methods to operate on the Trie's key set (retrieve, count, prefix search)

### DIFF
--- a/src/main/java/sirius/kernel/commons/Trie.java
+++ b/src/main/java/sirius/kernel/commons/Trie.java
@@ -16,7 +16,6 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
-import java.util.TreeSet;
 
 /**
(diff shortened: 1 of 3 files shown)


src/main/java/sirius/kernel/commons/Trie.java
```diff
@@ -329,4 +331,65 @@ public void put(@Nonnull CharSequence key, @Nonnull V value) {
...
+            result.addAll(getAllKeysBeginningWith(prefix + possibility, iter));
+            iter.goBack();
+        }
+
+        return result;
+    }
+
+    @Override
+    public String toString() {
+        return Strings.join(new TreeSet<>(keySet()), ", ");
```
andyHa: isn't that a bad idea, considering that Tries easily collect 10.000s of entries...?

D3nnisH: in this case, it is indeed

PR #84, opened 2017-05-26
Makes it very easy to mock Parts when testing

D3nnisH:
Simply extend the class to be mocked an Register your new class as a `TestMock` class:
Said, you want to mock a class `Converter`, create the following class:

```
@Register(classes = {Converter.class, TestMock.class})
public class ConverterMock extends Converter implements TestMock {

    /**
     * doin' nothin' special
     */
    @Override
    public Result execute(Param param) {
        return 42;
    }
}
```

This injects `ConverterMock` to all places, where
```
@Part
private Converter converter;
```
would usually inject the original `Converter` class, but only when running in TEST mode (aka when calling `TestHelper.setUp()` from you TestSuite)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,7 @@
         <version>3.9</version>
     </parent>
     <artifactId>sirius-kernel</artifactId>
-    <version>6.8</version>
+    <version>6.10</version>
     <name>Sirius Kernel</name>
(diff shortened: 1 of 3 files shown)


D3nnisH:
already implemented by `@Replace` 😅

PR #88, opened 2017-08-04
Uses NLS method to parse a string representing a boolean value

tbiScireum:
This ensures a correct handling of translated string like "Ja" or "Nein"


src/main/java/sirius/kernel/commons/Value.java
```diff
@@ -413,7 +413,7 @@ public Value first() {
             if (Strings.isEmpty(data)) {
                 return (T) Boolean.FALSE;
             }
-            return (T) (Boolean) Boolean.parseBoolean(String.valueOf(data));
+            return (T) NLS.parseUserString(Boolean.class, String.valueOf(data));
```
tbiScireum: done

PR #91, opened 2017-08-21
Respects decimalFormat while reading int/long Values from Strings

meg1502:


andyHa:
The changes which caused this fix were reverted...

PR #96, opened 2017-09-13
Sets the requestMethod to POST if postData is called

tbiScireum:


src/main/java/sirius/kernel/xml/Outcall.java
```diff
@@ -89,12 +83,14 @@ public Outcall(URL url, Context params, Charset charset) throws IOException {
...
      * Sents the given context as POST to the designated server.
      *
-     * @param params the data to POST
+     * @param params  the data to POST
      * @param charset the charset to use when encoding the post data
      * @return the outcall itself for fluent method calls
      * @throws IOException in case of any IO error
      */
     public Outcall postData(Context params, Charset charset) throws IOException {
+        connection.setRequestMethod("POST");
```
andyHa: line 82....

PR #117, opened 2018-03-01
Transforming also checks superclass

fkrone:
In some cases there might be a need to transform two classes which are not directly bound by a transformer e.g. if the source class is a more specific implementation. In this case it was needed to write multiple transformers in order to support this case. With these changes not only the source class itself is checked for transforming, but also its superclass. If this is not desired, the transformer can indicate this by setting a flag.

### DIFF
--- a/src/main/java/sirius/kernel/di/transformers/Transformer.java
+++ b/src/main/java/sirius/kernel/di/transformers/Transformer.java
@@ -35,6 +35,18 @@ public interface Transformer<S, T> extends Priorized {
     }
 
     /**
+     * Determines if child classes of the provided source class are also supported by this transformer.
+     * <p>
+     * In order to avoid unwanted transformations, this can be set to false. In this case, only the matching
+     * class to the provided source class is transformed.
+     *
+     * @return <tt>true</tt> if child classes should also be transformed via this transformer, <tt>false</tt> otherwise
+     */
+    default boolean supportChildClasses() {
+        return true;
+    }
+
+    /**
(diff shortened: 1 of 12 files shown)


src/main/java/sirius/kernel/di/transformers/Transformer.java
```diff
@@ -34,6 +34,18 @@ default int getPriority() {
...
 
+    /**
+     * Determines if child classes of the provided source class are also supported by this transformer.
+     * <p>
+     * In order to avoid unwanted transformations, this can be set to false. In this case, only the matching
+     * class to the provided source class is transformed.
+     *
+     * @return <tt>true</tt> if child classes should also be transformed via this transformer, <tt>false</tt> otherwise
+     */
+    default boolean supportChildClasses() {
```
andyHa: is that needed?

andyHa: *reuqired

src/main/java/sirius/kernel/di/transformers/Transformers.java
```diff
@@ -64,14 +69,28 @@
      */
     @SuppressWarnings("unchecked")
     public <T> T make(Object source, Class<T> target) {
-        for (Transformer<?, ?> adapterFactory : getFactories(source.getClass(), target)) {
+        T firstTransformResult = makeWithClass(source, source.getClass(), target, false);
```
andyHa: shouldn't that be recursive?

fkrone:
New implementation

PR #122, opened 2018-03-23
Fixes bug in eviction of caches

welzfabian:
It actually removed every CacheEntry which did NOT reach its ttl, not the other way around

### DIFF
--- a/src/test/java/sirius/kernel/cache/ManagedCacheSpec.groovy
+++ b/src/test/java/sirius/kernel/cache/ManagedCacheSpec.groovy
@@ -1,0 +1,32 @@
+/*
+ * Made with all the love in the world
+ * by scireum in Remshalden, Germany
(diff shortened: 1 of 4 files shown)


src/test/java/sirius/kernel/cache/ManagedCacheSpec.groovy
```diff
@@ -0,0 +1,32 @@
...
+
+class ManagedCacheSpec extends BaseSpecification {
+
+    def "test run eviction removes old entries"() {
+        given:
+        def cache = new ManagedCache("test-cache", null, null)
+        when:
+        cache.put("key1", "value1")
+        cache.put("key2", "value2")
+        Wait.seconds(11)
```
andyHa: does this have to be that long - builds are slow enough already

PR #124, opened 2018-03-26
SI-Prefixes

jakobvogel:
* Fixes file size formatter to yield precise SI-prefixed values.
* **Process before #125**


andyHa:
please incerement pom.xml - version by 0.0.1 or 0.1

PR #128, opened 2018-04-10
Fixes the Translation file Load order

tbiScireum:
- Fixes the customization resource recognition

Up to now all the customization translation files were not recognized as customization files. This lead to a random load order of the translation files in the babelfish class.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,7 @@
         <version>3.11.4</version>
     </parent>
     <artifactId>sirius-kernel</artifactId>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -9,7 +9,7 @@
         <version>3.11.4</version>
     </parent>
     <artifactId>sirius-kernel</artifactId>
-    <version>8.0.4</version>
+    <version>8.0.5</version>
```
andyHa: 9.0.0 - breaking change

PR #129, opened 2018-04-18
Adds a simple helper method to delete a single file which logs on failure

meg1502:


src/main/java/sirius/kernel/commons/Files.java
```diff
@@ -22,6 +26,8 @@
  */
 public class Files {
 
+    protected static final Log LOG = Log.get("files");
```
andyHa: no we need a specific logger? this is essentially a "cannot happen" case -> handle via default logger

src/main/java/sirius/kernel/commons/Files.java
```diff
@@ -137,6 +143,23 @@ public static String getFilenameWithoutExtension(@Nullable String path) {
...
     }
 
+    /**
+     * If the given file is not null and exists, tries to delete that file and logs when a file cannot be deleted. This is useful for error reporting and to
+     * diagnose why a file cannot be deleted.
+     *
+     * @param file the file to delete
+     */
+    public static void delete(@Nullable File file) {
+        if (file == null) {
```
andyHa: one of the rare cases where I'd invert the if

PR #135, opened 2018-06-07
Adds a Method to Limit to check how many Items remain to be processed

meg1502:
This is useful as upperLimit for the actual querys using this Limit

### DIFF
--- a/src/main/java/sirius/kernel/commons/Limit.java
+++ b/src/main/java/sirius/kernel/commons/Limit.java
@@ -135,6 +135,17 @@ public class Limit {
         return maxItems == 0 ? 0 : itemsToSkip + maxItems;
     }
 
+    /**
+     * Returns the remaining number of items to be processed. This is the number of items to be skipped plus the
+     * number of items to be accepted minus the number of items already processed.
+     * Returns <tt>null</tt> to indicate that there is no upper limit.
+     *
+     * @return the remaining number of items to be processed or <tt>null</tt> if there is no upper limit
+     */
+    public Integer getRemainingItems() {
+        return itemsToOutput == null ? null : itemsToSkip + itemsToOutput;
+    }
+
     @Override
     public String toString() {
         StringBuilder sb = new StringBuilder("Limit: ");
--- a/src/test/java/sirius/kernel/commons/LimitTest.java
+++ b/src/test/java/sirius/kernel/commons/LimitTest.java
@@ -14,8 +14,7 @@ import org.junit.Test;
 import java.util.List;
 import java.util.stream.Stream;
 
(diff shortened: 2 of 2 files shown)


src/main/java/sirius/kernel/commons/Limit.java
```diff
@@ -135,6 +135,17 @@ public int getTotalItems() {
...
     }
 
+    /**
+     * Returns the remaining number of items to be processed. This is the number of items to be skipped plus the
+     * number of items to be accepted minus the number of items already processed.
+     * Returns <tt>null</tt> to indicate that there is no upper limit.
+     *
+     * @return the remaining number of items to be processed or <tt>null</tt> if there is no upper limit
+     */
+    public Integer getRemainingItems() {
```
andyHa: int and default to 0 ?

src/main/java/sirius/kernel/commons/Limit.java
```diff
@@ -135,6 +135,17 @@ public int getTotalItems() {
...
+        return itemsToOutput == null ? null : itemsToSkip + itemsToOutput;
```
andyHa: what happens for Limit (5,0)

src/test/java/sirius/kernel/commons/LimitTest.java
```diff
@@ -86,4 +85,22 @@ public void testPredicate() {
         int sum = Stream.of(1, 2, 3, 4, 5, 6, 7, 8, 9, 10).filter(new Limit(2, 4).asPredicate()).mapToInt(i -> i).sum();
         assertEquals(3 + 4 + 5 + 6, sum);
     }
+
+    @Test
+    public void testRemaingItems() {
+        Limit l = new Limit(5, 500);
+        assertEquals(Integer.valueOf(505), l.getRemainingItems());
+        l = new Limit(10, 0);
+        assertNull(l.getRemainingItems());
```
andyHa: shouldn't this be 10?

meg1502: Limits created with maxItems < 1 are unlimited

src/main/java/sirius/kernel/commons/Limit.java
meg1502: returns null because Limits created with maxItems < 1 are unlimited

src/main/java/sirius/kernel/commons/Limit.java
meg1502: in contrast to getTotalItems, if enough items are processed, this method can actually return 0

so if it was changed to int returning 0 could mean either

a) unlimited items
b) all items already processed 

thats why i think returning null for the unlimited case is better

PR #137, opened 2018-06-13
Aha/better life

andyHa:


D3nnisH:
shouldn't we offer a possibility to escape the `|` symbol in counted translations? For example if I wanted to use `zero \| 0 | one \| 1 | many \| ${count}`

andyHa:
@D3nnisH let's implement this once we need it - so far we have no need and never had for placing an | in such a property

andyHa:
BUILD

src/test/resources/test_en.properties
```diff
@@ -0,0 +1,2 @@
+nls.test.withThree=zero|one|many: ${count}
```
meg1502: i dont think its very good to save 3 different translations under a single key.
since the versions are always the same wouldn't it be better to store them like this:

```
nls.test.withThree.zero=zero
nls.test.withThree.one=one
nls.test.withThree.many=many: ${count}
```

and make NLS fetch them using these suffixes?
Storing 3 different Translations does seem like bad practice to me and i can see problems when exporting such translations or confusion for the people translating these strings.

andyHa: this pattern is shamelessly stolen from other frameworks and it is almost a best practice to use one string...

src/main/java/sirius/kernel/settings/Settings.java
```diff
@@ -137,11 +143,23 @@ public Config getConfig(String key) {
...
             }
         }
 
+        result.sort((a, b) -> {
+            int prioA = a.hasPath(PRIORITY) ? a.getInt(PRIORITY) : Priorized.DEFAULT_PRIORITY;
+            int prioB = b.hasPath(PRIORITY) ? b.getInt(PRIORITY) : Priorized.DEFAULT_PRIORITY;
+
+            if (prioA == prioB) {
+                prioA = a.origin().lineNumber();
+                prioB = b.origin().lineNumber();
```
D3nnisH: smaller priority number means higher priority, gell? but shouldn't higher line number mean higher priority?

D3nnisH: indertat nicht!

PR #141, opened 2018-06-28
Extends Amount and StructuredOutput each by a helpful method

tbiScireum:


src/main/java/sirius/kernel/commons/Amount.java
```diff
@@ -635,6 +635,22 @@ public Amount negate() {
         return times(MINUS_ONE);
     }
 
+
+    /**
+     * Returns a {@link Amount} whose value is {@code (this % other)}.
```
Tobsucht: i think "mod" instead of "%" is a little bit clearer

src/test/java/sirius/kernel/commons/AmountSpec.groovy
```diff
@@ -341,4 +340,21 @@ class AmountSpec extends BaseSpecification {
         Amount.ZERO        | Amount.ZERO
         Amount.NOTHING     | Amount.NOTHING
     }
+
+    def "remainder() works as expected"() {
```
Tobsucht: test negative values?

src/main/java/sirius/kernel/xml/AbstractStructuredOutput.java
```diff
@@ -275,4 +275,12 @@ public StructuredOutput nullsafeProperty(@Nonnull String name, @Nullable Object
         property(name, data != null ? data : "");
         return this;
     }
+
+    @Override
+    public StructuredOutput propertyIfFilled(@Nonnull String name, @Nullable Object data) {
+        if (data != null) {
```
andyHa: whats the point of that? using property: null is more obvious and should yield the same result

PR #143, opened 2018-07-02
Total count for average

fkrone:
The total count can be used to determine the amount of values added to the average class.

Tags: average


src/main/java/sirius/kernel/health/Average.java
```diff
@@ -73,6 +74,7 @@ public void addValue(double value) {
      */
     public void addValues(long numberOfValues, double sumOfValue) {
         long newCount = count.addAndGet(numberOfValues);
+        totalCount.addAndGet(numberOfValues);
```
andyHa: what is the overflow behaviour - should be somehow reset it before Long.MAX_VALUE is reached?

PR #147, opened 2018-07-17
Allows NLS.parseMachineString to parse to an Amount

jmuscireum:


src/test/java/sirius/kernel/nls/NLSSpec.groovy
```diff
@@ -168,6 +168,16 @@ class NLSSpec extends BaseSpecification {
         "0.1" | new BigDecimal("0.1")
     }
 
+    def "parseMachineString works for Amount"() {
+        expect:
+        NLS.parseMachineString(Amount.class, input) == output
+
+        where:
+        input | output
+        "0.1" | Amount.ONE.divideBy(Amount.TEN)
```
meg1502: Add a test for empty String and null which are way more intresting then these basic numbers.

src/main/java/sirius/kernel/nls/NLS.java
```diff
@@ -1039,6 +1039,13 @@ private static String formatSpokenRecentDateWithTime(LocalDateTime givenDateTime
                 throw new IllegalArgumentException(fmtr("NLS.errInvalidDecimalNumber").set("value", value).format(), e);
             }
         }
+        if (Amount.class.equals(clazz)) {
```
meg1502: isn't this super confusing that null or "" return null instead of Amount.NONE?

src/test/java/sirius/kernel/nls/NLSSpec.groovy
jmuscireum: done

src/main/java/sirius/kernel/nls/NLS.java
jmuscireum: You are right, the alternative would be a workaround in sirius-search

PR #148, opened 2018-07-18
Adds a possibility for Handlers to interrupt XMLReader parsing

meg1502:
This can be useful when verifying a header or generally reading elements that are usually found near the start of the document.

### DIFF
--- a/src/main/java/sirius/kernel/xml/XMLReader.java
+++ b/src/main/java/sirius/kernel/xml/XMLReader.java
@@ -150,6 +150,14 @@ public class XMLReader extends DefaultHandler {
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/kernel/xml/XMLReader.java
```diff
@@ -149,6 +149,14 @@ public void parse(InputStream stream) throws IOException {
         private static final long serialVersionUID = -7454219131982518216L;
     }
 
+    /**
```
andyHa: do we really need 2 distinct exceptions for the exact same purpose?

why not wrap processing in its own task context and then simply call taskContext.cancel();

PR #160, opened 2018-10-16
Does not execute all the tests twice when using the ScenarioSuite

D3nnisH:
# before

image

# after

image


andyHa:
But we want to execute the main scenario!

PR #166, opened 2018-11-06
Adds method to provide lang to a CallContext without overriding alrea…

meg1502:
…dy set lang

### DIFF
--- a/src/main/java/sirius/kernel/async/CallContext.java
+++ b/src/main/java/sirius/kernel/async/CallContext.java
@@ -402,7 +402,7 @@ public class CallContext {
      *
      * @param lang the two-letter language code for this thread.
      */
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/kernel/async/CallContext.java
```diff
@@ -394,6 +394,20 @@ public void setLang(@Nullable String lang) {
...
 
+    /**
+     * Sets the current language for the current thread, only if no other language has been set already.
+     * <p>
+     * If <tt>null</tt> or an empty string is passed in, the language will not be changed.
+     * </p>
+     *
+     * @param lang the two-letter language code for this thread.
+     */
+    public void provideLang(@Nullable String lang) {
```
andyHa: setLangIfEmpy ?

PR #173, opened 2018-12-11
Won't log any exceptions which where caused by handledExceptions

tbiScireum:
This happens e.g. when we render pdfs. The external library will add our handledException to the root
cause but will throw its own exception. We would then handle the same exception twice, which is unnecessary.
- Fixes: SE-5174

### DIFF
--- a/src/main/java/sirius/kernel/health/Exceptions.java
+++ b/src/main/java/sirius/kernel/health/Exceptions.java
@@ -194,7 +194,6 @@ public class Exceptions {
                 processError = false;
                 LOG.FINE("Did not process the exception %s because its root (%s) was already handled",
                         ex.getMessage(), Exceptions.getRootCause(ex).getMessage());
-
             }
 
             try {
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/kernel/health/Exceptions.java
```diff
@@ -188,6 +189,14 @@ public HandledException handle() {
             if (ex instanceof HandledException) {
                 return (HandledException) ex;
             }
+
+            if (Exceptions.getRootCause(ex) instanceof HandledException) {
+                processError = false;
+                LOG.FINE("Did not process the exception %s because its root (%s) was already handled",
+                        ex.getMessage(), Exceptions.getRootCause(ex).getMessage());
+
```
andyHa: remove blank line

src/main/java/sirius/kernel/health/Exceptions.java
```diff
@@ -392,9 +401,29 @@ public static void logDeprecatedMethodUse() {
...
+    }
+    
+    /**
+     * Retrieves the actual root {@link Throwable} which ended in the given exception.
+     *
+     * @param e the throwable to begin with
+     * @return the root {@link Throwable} of the given one
+     */
+    public static Throwable getRootCause(@Nullable Throwable e) {
+        if (e == null){
```
andyHa: run code formatter

src/main/java/sirius/kernel/health/Exceptions.java
```diff
@@ -392,9 +401,29 @@ public static void logDeprecatedMethodUse() {
...
+            return null;
+        }
+
+        Throwable cause = e;
+
+        while (cause.getCause() != null) {
```
andyHa: some exceptions may have themselve as cause (I think..)

Maybe check for that and maybe also add a circuit breaker like "give up after 10 iterations"..

PR #175, opened 2019-01-11
Adds easier access to attributes from a StructredNode

meg1502:


src/main/java/sirius/kernel/xml/StructuredNode.java
```diff
@@ -143,6 +149,65 @@ public Node getNode() {
...
+     *
+     * @return the value of the attribute as a {@link String} if an attribute exists for the given name, <tt>""</tt> otherwise
+     */
+    @Nonnull
+    public String getAttributeValue(String name) {
+        StructuredNode attribute = getAttribute(name);
+        if (attribute != null) {
+            return attribute.getNode().getNodeValue();
+        }
+        return "";
```
meg1502: Or should null be returned in this case?

sabieber: Maybe an Optional?

src/main/java/sirius/kernel/xml/StructuredNode.java
```diff
@@ -143,6 +149,65 @@ public Node getNode() {
         return resultList;
     }
 
+    /**
+     * Returns a map of all attributes of this DOM node with their names as keys.
+     *
+     * @return a list containing all attributes of this node. If no attributes exist, an empty list will be returned.
```
welzfabian:
```suggestion
     * @return a map containing all attributes of this node. If no attributes exist, an empty map will be returned.
```

src/main/java/sirius/kernel/xml/StructuredNode.java
```diff
@@ -143,6 +149,65 @@ public Node getNode() {
...
+     */
+    @Nonnull
+    public ImmutableMap<String, StructuredNode> getAttributes() {
```
welzfabian: wouldn't it be the best practice to use `Map<...>` (the interface) instead of `ImmutableMap<...>` (a more specific implementation)?

welzfabian: ok, for ImmutableMaps this seems to be a more complicated issue...
https://stackoverflow.com/questions/38087900/is-it-better-to-return-an-immutablemap-or-a-map 

you can leave it like that, in my opinion

andyHa: Collections.unmodifiableMap()?

src/main/java/sirius/kernel/xml/StructuredNode.java
```diff
@@ -143,6 +149,65 @@ public Node getNode() {
...
+        nodeAttributes = attributes;
+    }
+
+    /**
+     * Returns the attribute with the given name.
+     *
+     * @return the attribute as a {@link StructuredNode} if an attribute exists for the given name, <tt>null</tt> otherwise
+     */
+    @Nullable
+    public StructuredNode getAttribute(String name) {
```
andyHa: why not return a Value here?

src/main/java/sirius/kernel/xml/StructuredNode.java
```diff
@@ -143,6 +149,65 @@ public Node getNode() {
...
+        return getAttributes().get(name);
```
andyHa: isn't it overkill to compute the whole map here?

src/main/java/sirius/kernel/xml/StructuredNode.java
```diff
@@ -143,6 +149,65 @@ public Node getNode() {
...
+    }
+
+    /**
+     * Returns the value of the  attribute with the given name.
+     *
+     * @return the value of the attribute as a {@link String} if an attribute exists for the given name, <tt>""</tt> otherwise
+     */
+    @Nonnull
+    public String getAttributeValue(String name) {
+        StructuredNode attribute = getAttribute(name);
```
andyHa: s.a.

src/main/java/sirius/kernel/xml/StructuredNode.java
andyHa: Optional or simply just use Value as stated above...

src/main/java/sirius/kernel/xml/StructuredNode.java
```diff
@@ -143,6 +149,65 @@ public Node getNode() {
...
+        return "";
+    }
+
+    /**
+     * Returns a map of all attribute values of this DOM node with their names as keys.
+     *
+     * @return a list containing all attribute values of this node. If no attributes exist, an empty list will be returned.
+     */
+    @Nonnull
+    public Map<String, String> getAttributeValues() {
```
andyHa: use this as above...

src/main/java/sirius/kernel/xml/StructuredNode.java
```diff
@@ -143,6 +147,37 @@ public Node getNode() {
...
+    }
+
+    /**
+     * Returns the value of the attribute with the given name.
+     *
+     * @return a {@link Value} filled with the attribute value if an attribute exists for the given name, an empty {@link Value} otherwise.
+     */
+    @Nonnull
+    public Value getAttribute(String name) {
+        StructuredNode attribute = queryNode("@" + name);
```
andyHa:
```suggestion
        if (node.getAttributes() != null) {
            Node attribute = node.getAttributes().getNamedItem(name);
            if (attribute != null) {
               return Value.of(attribute.getNodeValue());
            }
         }      
```

PR #176, opened 2019-01-21
Aha/sonarlint

andyHa:


src/main/java/sirius/kernel/di/PartRegistry.java
```diff
@@ -269,11 +269,7 @@ private void registerNamedPart(String uniqueName,
                                    String customizationName,
                                    Class<?> clazz,
                                    Map<String, Object> partsOfClass) {
-        Object currentPart = partsOfClass.get(uniqueName);
-        if (currentPart == null) {
-            partsOfClass.put(uniqueName, part);
-            return;
-        }
+        Object currentPart = partsOfClass.computeIfAbsent(uniqueName, name -> part);
```
D3nnisH: behaviour has changed: the `return` is missing

andyHa: args - thx!!

PR #192, opened 2019-04-15
Adds the ability to start programs in different working directories

meg1502:
Previously, every external program started via Exec inherited its working directory from the current process. With the introduction of a *direcotry* parameter, this can now be changed.

This is a non-breaking change, and calls to the old method signature will behave the same as before.

Fixes: OX-5188

### DIFF
--- a/src/main/java/sirius/kernel/commons/Exec.java
+++ b/src/main/java/sirius/kernel/commons/Exec.java
@@ -14,6 +14,7 @@ import sirius.kernel.health.Log;
 import sirius.kernel.nls.NLS;
 
 import java.io.BufferedReader;
+import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/kernel/commons/Exec.java
```diff
@@ -146,9 +147,25 @@ public static String exec(String command, boolean ignoreExitCodes) throws ExecEx
...
+     *
+     * @param command         the command to execute
+     * @param ignoreExitCodes if an exit code other than 0 should result in an exception being thrown
+     * @param opTimeout       the duration after which the execution should be logged as hanging
+     * @param directory       the working directory of the subprocess,
+     *                        or <tt>null</tt> if the subprocess should inherit the working directory of the current process.
+     * @return the transcript of stderr and stdout produced by the executed command
+     * @throws ExecException in case the external program fails
+     */
+    public static String exec(String command, boolean ignoreExitCodes, Duration opTimeout, File directory)
```
andyHa: @Nullable File (can't suggest as import would be missing)

PR #202, opened 2019-07-03
 Add resetLang to be able to retrigger the recalculation of the language

welzfabian:


src/main/java/sirius/kernel/async/CallContext.java
```diff
@@ -427,6 +427,13 @@ public void deferredSetLang(@Nonnull Consumer<CallContext> languageInstaller) {
         this.lazyLanguageInstaller = languageInstaller;
     }
 
+    /**
+     * Sets the lang back to null. Use with caution.
```
andyHa: Why?

andyHa: Sets the lang back to null.

This method should only be used to re-initialize the language. Use ...#setLang... to specify a new language.

PR #214, opened 2019-08-30
Permits to control and observe the execution of background loops with…

andyHa:
…in tests.

Many tests need to control or observe if and when a loop is executed - therefore
a generic API is provided.

### DIFF
--- a/src/main/java/sirius/kernel/async/BackgroundLoop.java
+++ b/src/main/java/sirius/kernel/async/BackgroundLoop.java
@@ -9,7 +9,9 @@
 package sirius.kernel.async;
 
 import sirius.kernel.Sirius;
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/kernel/async/BackgroundLoop.java
```diff
@@ -173,4 +201,91 @@ public String getExecutionInfo() {
     public Instant getLastExecutionAttempt() {
         return Instant.ofEpochMilli(lastExecutionAttempt);
     }
+
+    /**
+     * Returns the execution future of the given background loop
```
meg1502:
```suggestion
     * Returns the execution future of the given background loop.
```

PR #219, opened 2019-09-18
Improves converting Amounts inside Values into BigDecimals

meg1502:


src/main/java/sirius/kernel/commons/Value.java
```diff
@@ -1173,6 +1173,9 @@ public BigDecimal getBigDecimal() {
             if (data instanceof BigDecimal) {
                 return (BigDecimal) data;
             }
+            if (data instanceof Amount) {
+                return ((Amount) data).getAmount();
+            }
             if (data instanceof Double) {
```
andyHa: add the same to getAmount? if it is already one why wrap again?

PR #220, opened 2019-09-23
Adds flatMapping from Values to Optionals

meg1502:
This for example can simplify code like this:

```
portal.getConfig("standard.requestTargetPortal")
                                    .map(value -> mango.find(Portal.class, value.asString()).orElse(portal))
                                    .orElse(portal);
```

to

```
portal.getConfig("standard.requestTargetPortal")
                                    .flatMap(value -> mango.find(Portal.class, value.asString()))
                                    .orElse(portal);
```

and can save even more depending on the onElse implementation

### DIFF
--- a/src/main/java/sirius/kernel/commons/Value.java
+++ b/src/main/java/sirius/kernel/commons/Value.java
@@ -250,9 +250,10 @@ public class Value {
 
     /**
      * Returns an optional value computed by the given mapper. If this value is <tt>empty</tt> the mapper will not
-     * be called, but an empty optional will be returned. This method is similar to {@link #map(Function)},
-     * but the provided mapper is one whose result is already an {@link Optional}, and if invoked,
-     * {@code flatMap} does not wrap it with an additional {@link Optional}.
+     * be called, but an empty optional will be returned.
+     * <p>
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/kernel/commons/Value.java
```diff
@@ -249,6 +248,25 @@ public Value replaceIfEmpty(@Nonnull Supplier<?> supplier) {
         }
     }
 
+    /**
+     * Returns an optional value computed by the given mapper. If this value is <tt>empty</tt> the mapper will not
```
andyHa: new line / p tag

src/main/java/sirius/kernel/commons/Value.java
```diff
@@ -249,6 +248,25 @@ public Value replaceIfEmpty(@Nonnull Supplier<?> supplier) {
...
+     * be called, but an empty optional will be returned. This method is similar to {@link #map(Function)},
```
andyHa:
```suggestion
     * be called, but an empty optional will be returned.
     * <p>
     * This method is similar to {@link #map(Function)},
```

PR #224, opened 2019-10-10
Aha/optimistic locking helper

andyHa:


src/main/java/sirius/kernel/commons/OptimisticLockAlgorithm.java
```diff
@@ -0,0 +1,116 @@
...
+     * @param cleanup         the callback which undoes the changes of <tt>factory</tt> if a collision was detected by
+     *                        <tt>correctnessTest</tt>
+     * @return the created object or an empty optional if the algorithm ran out of retries
+     * @throws Exception in case any of the provided callbacks threw an exception
+     */
+    public Optional<R> tryToCreate(Producer<R> factory, Processor<R, Boolean> correctnessTest, Callback<R> cleanup)
+            throws Exception {
+
+        int retries = numberOfRetries;
+        while (retries-- > 0) {
```
meg1502: This should be >= 0 or even < 0 ; depending if retries is really the number of "retries" or simply the number of "attempts".

Currently if i have retries = 5 there will only be 4 attempts as the first attempt already decreases the variable to 4 and on the 5. iteration 0 > 0 is false

meg1502: or do it like this:
```
        int retries = numberOfRetries;
        while (retries > 0) {
            retries--;
```

andyHa: while(retries-- > 0) ... and wile(retries > 0) ... retires---; is the same

PR #230, opened 2019-11-26
Splits a overly complex method into distinct method calls.

andyHa:


andyHa:
Keep open until post 2019.12

PR #231, opened 2019-12-05
SIRI-127 Prevents a NPE when trying to cast a Value

jmuscireum:
get(defaultValue) will return null if defaultValue is null and the value is empty. This will cause a NPE when calling asOptional on an empty value.

### DIFF
--- a/src/main/java/sirius/kernel/commons/Value.java
+++ b/src/main/java/sirius/kernel/commons/Value.java
@@ -317,7 +317,7 @@ public class Value {
      */
     @Nonnull
     public <T> Optional<T> asOptional(@Nonnull Class<T> type) {
-        return Optional.ofNullable(get(type, null));
+        return Optional.ofNullable(get(type));
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/kernel/commons/Value.java
```diff
@@ -604,14 +604,33 @@ public Value first() {
...
+     * @param clazz the desired class of the return type
+     * @param <V>   the expected type of the wrapped value
+     * @return the wrapped value if the given <tt>clazz</tt> is assignable from wrapped values class or <tt>null</tt>
+     * otherwise
+     */
+    @Nullable
+    @SuppressWarnings("unchecked")
+    public <V> V get(Class<V> clazz) {
+        Object result = get();
+        if (result == null || !clazz.isAssignableFrom(result.getClass())) {
```
meg1502: why not just add the nullcheck to the old method?

sabieber: nope, this is as intended. when you provide a default value you dont want the returning value to be null, so passing null would go against the point of the API. To provide methods where null as a return type is wanted and accepted a new method is added.

PR #232, opened 2019-12-05
Aha/value cleanup

andyHa:


andyHa:
This contains at least two breaking changes -> we have to release a major version once this is merged.

src/main/java/sirius/kernel/commons/Value.java
```diff
@@ -459,8 +459,7 @@ public Object get() {
      * @param defaultValue the value to use if the inner value is <tt>null</tt>
      * @return the wrapped value or the given defaultValue if the wrapped value is <tt>null</tt>
      */
-    @Nonnull
-    public Object get(@Nonnull Object defaultValue) {
+    public Object get(Object defaultValue) {
```
sabieber: Add Nullable annotation to param and return value for all methods where annotations are now removed?

src/main/java/sirius/kernel/commons/Value.java
```diff
@@ -609,29 +608,10 @@ public Value first() {
...
-     *
-     * @param clazz the desired class of the return type
-     * @param <V>   the expected type of the wrapped value
-     * @return the wrapped value if the given <tt>clazz</tt> is assignable from wrapped values class or <tt>null</tt>
-     * otherwise
-     */
-    @Nullable
-    @SuppressWarnings("unchecked")
-    public <V> V get(Class<V> clazz) {
+    public <V> V get(Class<V> clazz, V defaultValue) {
```
sabieber: couldnt we retain the get(Class) method with just an overload of (Class, null) to not break as much code? my guess is that a lot of these calls could also be in templates where now breaking code is not as easy to detect.
And it wouldnt increase the complexity a lot

src/main/java/sirius/kernel/commons/Value.java
```diff
@@ -539,31 +539,36 @@ public Value first() {
...
+            return (T) asLocalDateTime((LocalDateTime) defaultValue);
+        }
+        if (ZonedDateTime.class.equals(targetClazz) && is(TemporalAccessor.class,
+                                                          Calendar.class,
+                                                          Date.class,
+                                                          java.sql.Date.class,
+                                                          Timestamp.class)) {
+            return (T) asZonedDateTime((ZonedDateTime) defaultValue);
+        }
+        if (LocalTime.class.equals(targetClazz) && data instanceof TemporalAccessor && is(TemporalAccessor.class,
```
sabieber: sorry if I am missing something but how could data be assignable to e.g. Timestamp or Calendar when its an instance of TemporalAccessor?

src/main/java/sirius/kernel/commons/Value.java
andyHa: this will make sonarlint force you to add extra null checks where it is guaranteed to be non-null e.g. get("X") cannot be null - but the annotation system isn't powerful enought to make this guarantee

src/main/java/sirius/kernel/commons/Value.java
sabieber: Merge readded the wrong instanceof check

PR #233, opened 2019-12-16
Matches exception handling in Settings getter, so none will throw errors 

meg1502:


meg1502:
will be incoperated into #234

PR #234, opened 2019-12-17
Aha/cleanup

andyHa:


src/main/java/sirius/kernel/commons/Value.java
```diff
@@ -535,30 +535,34 @@ public Value first() {
...
-        if (LocalTime.class.equals(targetClazz) && data instanceof TemporalAccessor) {
-            if (is(TemporalAccessor.class,
-                   Calendar.class,
-                   Date.class,
-                   java.sql.Date.class,
-                   Timestamp.class,
-                   Time.class)) {
-                return (T) asLocalTime((LocalTime) defaultValue);
-            }
+        if (LocalDate.class.equals(targetClazz) && is(TemporalAccessor.class,
```
sabieber: This commit seems to be a duplicate

PR #238, opened 2020-02-13
Permits lambdas to decalre concise exceptions.

andyHa:


andyHa:
Change doesn't make that much sense...

PR #240, opened 2020-02-15
Adds a helper class to wrap a timeout as boolean predicate.

andyHa:


src/main/java/sirius/kernel/commons/Timeout.java
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Wraps a timestamp into a simple predicate.
+ */
+public class Timeout {
+
+    private final long timestamp;
+
+    /**
+     * Creates a new timeout which is toggled after to given amount of time.
```
idlira:
```suggestion
     * Creates a new timeout which is toggled after the given amount of time.
```

src/main/java/sirius/kernel/commons/Timeout.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    /**
+     * Returns <tt>true</tt> as long as the timeout has not been reached.
+     *
+     * @return <tt>true</tt> if the specified amount of time has not elapsed yet since the creation of the timeout,
+     * <tt>false</tt> otherwise
+     */
+    public boolean notReached() {
+        return System.currentTimeMillis() < timestamp;
```
sabieber:
```suggestion
        return !isReached();
```

PR #243, opened 2020-03-04
Mini-Additions to URLBuilder

meg1502:


src/main/java/sirius/kernel/commons/URLBuilder.java
```diff
@@ -126,4 +128,25 @@ public URLBuilder addParameter(@Nonnull String key, @Nullable Object value, bool
...
+
+    /**
+     * Creates a {@link URL URL object} from the resulting string of this builder.
+     * <p>
+     * Only works if the url contains a valid protocol.
+     *
+     * @return the url that was built as a {@link URL URL object}
+     * @throws MalformedURLException should only happen if no protocol or an invalid protocol has been given
+     */
+    public URL asURL() throws MalformedURLException {
```
andyHa: wrap in an IllegalArgumentException or IllegalStateException as there is most probably no specific code to handle this exact exception - spares an extra try/catch for each time it is used.

meg1502:
sample error: `Could not create URL: no protocol: sirius-lib.net`

PR #247, opened 2020-03-21
Creates new method to mark the outcall as post request

tbiScireum:
- Fixes: SE-9304

### DIFF
--- a/src/main/java/sirius/kernel/xml/Outcall.java
+++ b/src/main/java/sirius/kernel/xml/Outcall.java
@@ -25,6 +25,7 @@ import java.io.OutputStreamWriter;
 import java.io.StringWriter;
 import java.io.UnsupportedEncodingException;
 import java.net.HttpURLConnection;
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/kernel/xml/Outcall.java
```diff
@@ -96,6 +97,17 @@ public Outcall postData(Context params, Charset charset) throws IOException {
         return this;
     }
 
+    /**
+     * Marks the request as POST request.
+     *
+     * @return the outcall itself for fluent method calls
+     * @throws ProtocolException if the method cannot be reset or if the requested method isn't valid for HTTP.
+     */
+    public Outcall markAsPostRequest() throws ProtocolException {
```
andyHa:
```suggestion
    public Outcall markAsPostRequest() throws IOException {
```

andyHa: and use this method in the code above (postData)

PR #248, opened 2020-03-30
Provides an annotation and a ClassLoadAction to synthesize transformers.

andyHa:
Many transformers simply invoke the constructor of the target class. Such
transformers can be synthesized by placing the AutoTransform annotation
on the target class which then specifies the source class of the
synthesized transformer.

### DIFF
--- a/src/main/java/sirius/kernel/di/transformers/AutoTransform.java
+++ b/src/main/java/sirius/kernel/di/transformers/AutoTransform.java
@@ -22,7 +22,7 @@ import java.lang.annotation.Target;
  * If a {@link Transformer} simply invokes the constructor of the target class and passes in the source instance,
  * this annotation can be used to synthesize such a transformer.
  * <p>
- * To achive this, the target class must wear this annotation and specify the source class as well as optionally
+ * To achieve this, the target class must wear this annotation and specify the source class as well as optionally
  * a priority (if another transformer is to be overwritten / bypassed).
  * <p>
  * Note that the annotated class has to provide a public single argument constructor which takes the source instance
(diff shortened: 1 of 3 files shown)


src/main/java/sirius/kernel/di/transformers/AutoTransform.java
```diff
@@ -0,0 +1,55 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Marks a class as automatically transformed.
+ * <p>
+ * If a {@link Transformer} simply invokes the constructor of the target class and passes in the source instance,
+ * this annotation can be used to synthesize such a transformer.
+ * <p>
+ * To achive this, the target class must wear this annotation and specify the source class as well as optionally
```
idlira:
```suggestion
 * To achieve this, the target class must wear this annotation and specify the source class as well as optionally
```

src/main/java/sirius/kernel/di/transformers/AutoTransform.java
```diff
@@ -0,0 +1,55 @@
...
+ * Note that no {@link sirius.kernel.di.std.Register} needs (and also should) be present next to this annotation.
+ *
+ * @see Transformer
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.TYPE, ElementType.METHOD})
+public @interface AutoTransform {
+
+    /**
+     * Specifies the source class to transform from int othe annotated class.
```
idlira:
```suggestion
     * Specifies the source class to transform from into the annotated class.
```

PR #249, opened 2020-03-31
Delays fetching the language as it isn't required in many cases.

andyHa:


src/main/java/sirius/kernel/nls/Formatter.java
```diff
@@ -149,6 +174,63 @@ public Formatter set(Map<String, Object> ctx) {
...
+     * @return <tt>this</tt> to permit fluent method chains
+     */
+    public Formatter setDirect(Map<String, Object> ctx) {
+        if (ctx != null) {
+            for (Map.Entry<String, Object> e : ctx.entrySet()) {
+                Object value = e.getValue();
+                if (value == null) {
+                    setDirect(e.getKey(), "");
+                } else {
+                    setDirect(e.getKey(), e.getValue().toString());
```
bwiedmann:
```suggestion
                    setDirect(e.getKey(), value.toString());
```

PR #250, opened 2020-04-02
Cache remover builder

qw3ry:


src/main/java/sirius/kernel/cache/Cache.java
```diff
@@ -157,6 +157,8 @@
      */
     Cache<K, V> addRemover(@Nonnull String discriminator, @Nonnull BiPredicate<String, CacheEntry<K, V>> test);
 
+    <T> CacheRemoverBuilder<K, V, CacheEntry<K, V>> addRemover(@Nonnull String discriminator);
```
andyHa, line 160: the T can be killed (?) - it is unused..

src/main/java/sirius/kernel/cache/CacheRemoverBuilder.java
```diff
@@ -0,0 +1,15 @@
+package sirius.kernel.cache;
+
+import java.util.function.BiFunction;
+import java.util.function.BiPredicate;
+import java.util.function.Function;
+import java.util.function.Predicate;
+
+public interface CacheRemoverBuilder<K, V, T> {
+    <R> CacheRemoverBuilder<K, V, R> map(BiFunction<String, T, R> mapper);
```
andyHa, line 9: probably not needed - as far as i understand it right now...

src/main/java/sirius/kernel/cache/Cache.java
qw3ry: ups. Ja.... noch nicht aufgeräumt und so^^

src/main/java/sirius/kernel/cache/CacheRemoverBuilder.java
```diff
@@ -0,0 +1,15 @@
...
+
+    Cache<K, V> removeIf(BiPredicate<String, T> predicate);
+    <R> CacheRemoverBuilder<K, V, R> map(Function<T, R> mapper);
+
+    Cache<K, V> removeIf(Predicate<T> predicate);
```
andyHa, line 14: probably unused as the discriminator is missing (?)

src/main/java/sirius/kernel/cache/ManagedCache.java
```diff
@@ -341,6 +342,11 @@ public void removeIf(@Nonnull Predicate<CacheEntry<K, V>> predicate) {
         return this;
     }
 
+    @Override
+    public CacheRemoverBuilder<K, V, CacheEntry<K, V>> addRemover(@Nonnull String discriminator) {
```
andyHa, line 346: also add a second method addValueBasedRemove or the like, which directly operates on the value of the cache entry (which is what we do in 95% of the time)

src/main/java/sirius/kernel/cache/ManagedCacheRemoverBuilder.java
```diff
@@ -0,0 +1,46 @@
+package sirius.kernel.cache;
+
+import java.util.function.BiFunction;
+import java.util.function.BiPredicate;
+import java.util.function.Function;
+import java.util.function.Predicate;
+
+public class ManagedCacheRemoverBuilder<K, V, T> implements CacheRemoverBuilder<K, V, T> {
```
andyHa, line 8: can be set to package visibility (?)

src/main/java/sirius/kernel/cache/CacheRemoverBuilder.java
qw3ry: Its very little implementation overhead, and it does allow some more flexibility.

src/main/java/sirius/kernel/cache/CacheRemoverBuilder.java
qw3ry: Same as above. Its very little implementation overhead, and it does allow some flexibility, like a `removeIf(Objects::nonNull)`. The discriminator might have been used earlier.

However, it might be a good idea to force the api user to ignore a parameter, he might be tempted to forget the descriminator totally

src/main/java/sirius/kernel/cache/Cache.java
```diff
@@ -152,11 +152,35 @@
...
+
+    /**
+     * Convenience function for {@code addRemover(discriminator).map(CacheEntry::getValue)}.
+     *
+     * @param discriminator the name of the remover
+     * @return a {@link CacheRemoverBuilder} to build the remover
+     * @see #addRemover(String)
+     * @see #addRemover(String, BiPredicate)
+     */
+    default CacheRemoverBuilder<K, V, V> addValueBasedRemover(@Nonnull String discriminator) {
```
andyHa, line 180: Although certainly not wrong - IMHO default should mainly be used for legacy "fixes" and we should move this into the impl - even if this means that it'll be a bit redundant...

src/main/java/sirius/kernel/cache/CacheRemoverBuilder.java
```diff
@@ -0,0 +1,123 @@
...
+    <R> CacheRemoverBuilder<K, V, R> map(BiFunction<String, T, R> mapper);
+
+    /**
+     * Convenience method for {@link #map(BiFunction)} ignoring the test parameter.
+     *
+     * @param mapper The mapping function that is applied
+     * @param <R>    The new underlying type that replaces T
+     * @return a new {@link CacheRemoverBuilder} with the mapped underlying type
+     */
+    default <R> CacheRemoverBuilder<K, V, R> map(Function<T, R> mapper) {
```
andyHa, line 67: I'd rather move the impl into the underlying class

src/main/java/sirius/kernel/cache/CacheRemoverBuilder.java
```diff
@@ -0,0 +1,123 @@
...
+     */
+    CacheRemoverBuilder<K, V, T> filter(BiPredicate<String, T> predicate);
+
+    /**
+     * Convenience method for {@link #filter(BiPredicate)} ignoring the test parameter.
+     *
+     * @param predicate The predicate to test the objects
+     * @return A new {@link CacheRemoverBuilder} operating on the remaining objects
+     */
+    default CacheRemoverBuilder<K, V, T> filter(Predicate<T> predicate) {
```
andyHa, line 87: s.a.

src/main/java/sirius/kernel/cache/CacheRemoverBuilder.java
```diff
@@ -0,0 +1,123 @@
...
+     */
+    CacheRemoverBuilder<K, V, T> removeAlways(BiPredicate<String, T> predicate);
+
+    /**
+     * Convenience method for {@link #removeAlways(BiPredicate)} ignoring the test parameter.
+     *
+     * @param predicate The predicate to test the objects
+     * @return A new {@link CacheRemoverBuilder} operating on the remaining objects
+     */
+    default CacheRemoverBuilder<K, V, T> removeAlways(Predicate<T> predicate) {
```
andyHa, line 106: s.a.

src/main/java/sirius/kernel/cache/CacheRemoverBuilder.java
```diff
@@ -0,0 +1,123 @@
...
+     * This method terminates the builder. All objects that are not yet ignored via {@link #filter} or
+     * removed via {@link #removeAlways} are evaluated with the given predicate. If they match the predicate,
+     * they will be removed from the cache.
+     *
+     * @param predicate The predicate to test the objects
+     * @return The cache this {@link CacheRemoverBuilder} operates on
+     */
+    Cache<K, V> removeIf(BiPredicate<String, T> predicate);
+
+    default Cache<K, V> removeIf(Predicate<T> predicate) {
```
andyHa, line 120: s.a. + add some javadocs :)

src/main/java/sirius/kernel/cache/CacheRemoverBuilder.java
```diff
@@ -0,0 +1,123 @@
...
+     *
+     * @param predicate The predicate to test the objects
+     * @return A new {@link CacheRemoverBuilder} operating on the remaining objects
+     */
+    default CacheRemoverBuilder<K, V, T> removeAlways(Predicate<T> predicate) {
+        return removeAlways((string, value) -> predicate.test(value));
+    }
+
+    /**
+     * This method terminates the builder. All objects that are not yet ignored via {@link #filter} or
```
andyHa, line 111: excellent javadoc but the summary of the method would be "This method terminates the builder."

We commonly use a singe sentence in a single line for a short description - followed (if necessarry) by a longer description.

andyHa, line 111:
```suggestion
     * Removes all cache entries matching the give filter.
     * <p>
     * This method terminates the builder. All objects that are not yet ignored via {@link #filter} or
```

src/main/java/sirius/kernel/cache/ManagedCacheRemoverBuilder.java
```diff
@@ -0,0 +1,86 @@
...
+    private final String discriminator;
+
+    /**
+     * The combined mapping/filter function up to now.
+     * The {@link State} defines whether the object was already marked for removal via {@link #removeAlways} or
+     * marked as filtered via {@link #filter}. If it is marked as undecided, it should be processed further.
+     */
+    private final BiFunction<String, CacheEntry<K, V>, Tuple<T, State>> mapper;
+
+    private enum State {
```
andyHa, line 19: the impl. is IMHO fine as it is, just as an additional thought: couldn't we have also provided two special instances of subclasses which either always or never remove and return these as "dummies" (which then would ignore all successive calls to map / filter / remove) (?)

src/main/java/sirius/kernel/cache/ManagedCacheRemoverBuilder.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.function.BiFunction;
+import java.util.function.BiPredicate;
+
+class ManagedCacheRemoverBuilder<K, V, T> implements CacheRemoverBuilder<K, V, T> {
+    private final Cache<K, V> cache;
+    private final String discriminator;
+
+    /**
+     * The combined mapping/filter function up to now.
+     * The {@link State} defines whether the object was already marked for removal via {@link #removeAlways} or
```
andyHa, line 14: add a <p> inbetween the lines

src/test/java/sirius/kernel/cache/ManagedCacheSpec.groovy
```diff
@@ -75,4 +80,34 @@ class ManagedCacheSpec extends BaseSpecification {
         cache.get("C") == null
         cache.get("D") != null
     }
+
+    def "remover builder works as expected"() {
+        given:
+        ManagedCache<String, Tuple<String, String>> cache = new ManagedCache("test-cache", null, null)
+        // lambdas in groovy seem kind of ... awkward
```
andyHa, line 87: true - no need to comment - A LOT of things are awkward in groovy...

src/main/java/sirius/kernel/cache/CacheRemoverBuilder.java
```diff
@@ -0,0 +1,134 @@
...
+ *             .filter(Objects::nonNull)
+ *             .removeIf(Strings::equals)
+ * }</pre>
+ * This is equivalent to
+ * <pre>{@code
+ * CacheManager.<Tuple<String, Object>>createCoherentCache("my-cache")
+ *             .addRemover("my-remover", (test, entry) -> {
+ *                 Tuple<String, Object> tuple = entry.getValue();
+ *                 if (tuple.getFirst() == null) {
+ *                     return false;
```
jmuscireum, line 36:
```suggestion
 *                     return true;
```
Oder nicht?

src/main/java/sirius/kernel/cache/CacheRemoverBuilder.java
```diff
@@ -0,0 +1,134 @@
...
+ * <p>
+ * A CacheRemoveBuilder is created via {@link Cache#addRemover(String)}. It allows to build a method reference
+ * for a cache remover step-by-step. For details on the cache removal, see {@link Cache#addRemover(String, BiPredicate)}.
+ * <p>
+ * The type parameter T is the type of the object this {@link CacheRemoverBuilder} operates on. Initially,
+ * it is a {@link CacheEntry}, but it can be changed using the {@link #map} function.
+ *
+ * <h3>Usage example</h3>
+ * The cache remover is an intuitive, stream-like API that maps the cache entries, and marks them for removal if
+ * the fulfill the given conditions:
```
jmuscireum, line 20:
```suggestion
 * they fulfill the given conditions:
```

PR #256, opened 2020-04-20
BREAKING: Refactors the way @Registers works.

andyHa:
Previously @Register took only the list of interfaces and registered the
newly instantiated part for all of these. This lead to a pretty common
pitfall if the target to register for was the class itself, another class
or an interface implemented by a common parent class.

Therefore we now provide the marker annotation @AutoRegister which
can be placed on all interfaces or classes which should end up in
the MutableGlobalContext.registerPart call.

To register for an interface which is implemented by this part, simply
add @Register to the part and @AutoRegister to the interface
(same as before).

To also register to the part class itself, simply add @AutoRegister next
to @Register on the part itself.

To register for an interface implemented by a parent class (e.g.
subclasses of BizController in sirius-biz), simply use @Register
on the controller as the Controller interface (will) wear @AutoRegister).

To register for a class (e.g. BackgroundLoop) also simply use
@Reigster on the loop part, as BackgroundLoop will also wear an
@AutoRegister annotation.


src/main/java/sirius/kernel/di/std/RegisterLoadAction.java
```diff
@@ -38,21 +40,51 @@
...
-                Injector.LOG.FINE("Using the fallback method to determine which %s will be registered for: %s", clazz.getName());
+                Injector.LOG.FINE("Using the fallback method to determine which %s will be registered for: %s",
+                                  clazz.getName());
             }
         }
 
+        // Warn if the classes list can (and should) be omitted...
         if (registeredClasses.equals(detectedClasses)) {
             Injector.LOG.WARN(
                     "%s wears a @Register with a list of classes which is also auto-redected. Consider removing the classes list...",
```
idlira, line 90:
```suggestion
                    "%s wears a @Register with a list of classes which are already auto-detected. Consider removing the classes list...",
```

PR #263, opened 2020-05-18
Reworks namespace handling in XMLCalls and provides a simple SOAP client

andyHa:
Why not use the Java JDK client? Will be dropped out of the JDK soon
Why not use an off the shelve client? There is none which doesn't bring a dependency hell with it.
Also the API should be quite concise. Further more, once we switched to Java 9+, we can use the
new HTTPClient, which permits connection pooling + keepalive etc.


src/main/java/sirius/kernel/xml/BasicNamespaceContext.java
```diff
@@ -0,0 +1,72 @@
...
+     */
+    public BasicNamespaceContext withNamespace(String prefix, String uri) {
+        this.prefixToUri.put(prefix, uri);
+        return this;
+    }
+
+    /**
+     * Returns all registered prefixes along with their namespace URIs.
+     *
+     * @return a stream of entries mapping prefixes to their namespcae URIs
```
bwiedmann, line 42:
```suggestion
     * @return a stream of entries mapping prefixes to their namespace URIs
```

src/main/java/sirius/kernel/xml/XMLStructuredInput.java
```diff
@@ -30,24 +32,33 @@
...
      * Creates a new XMLStructuredInput for the given stream.
      *
-     * @param in    the InputStream containing the xml data.
-     * @param close determines whether the stream should be closed after parsing or not
+     * @param inputStream      the InputStream containing the xml data
+     * @param namespaceContext the namespace context to use when applying XPATH queries
      * @throws IOException if an io error occurs while parsing the input xml
      */
-    public XMLStructuredInput(InputStream in, boolean close) throws IOException {
+    public XMLStructuredInput(InputStream inputStream,@Nullable NamespaceContext namespaceContext) throws IOException {
```
bwiedmann, line 39: Code format?

PR #266, opened 2020-05-27
Drops log4j and reduces usage of Guava.

andyHa:
log4j doesn't provide any benefit for us anymore and only adds to the complexity of the system. Therefore it was removed and we internally use JULI - which is there anyway and also support SLF4J in case a library uses it.

This change should remain mostly unnoticed by users of sirius. The only breaking change
are the Level names WARN (old) -> WARNING (new) and DEBUG (old) -> FINE (new).

We also removed as many calls to Google Guava as possible. Most of the calls were only benificial in the pre Java 8 world (no diamond operator). Others more troublesome as Hashing and ByteStreams was marked as unstable and now partially as deprecated  therefore these parts have been replaced.

### DIFF
--- a/src/main/java/sirius/kernel/health/Log.java
+++ b/src/main/java/sirius/kernel/health/Log.java
@@ -8,12 +8,7 @@
 
 package sirius.kernel.health;
 
-import com.google.common.collect.Lists;
-import org.apache.log4j.Level;
-import org.apache.log4j.Logger;
-import org.apache.log4j.MDC;
 import sirius.kernel.Sirius;
-import sirius.kernel.async.CallContext;
 import sirius.kernel.commons.Explain;
 import sirius.kernel.commons.Strings;
 import sirius.kernel.di.PartCollection;
@@ -22,6 +17,12 @@ import sirius.kernel.nls.NLS;
 
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.logging.Level;
+import java.util.logging.LogRecord;
+import java.util.logging.Logger;
 
 /**
  * The logging facade used by the system.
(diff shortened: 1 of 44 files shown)


src/main/java/sirius/kernel/health/Log.java
```diff
@@ -131,86 +139,52 @@ public static synchronized Log get(String name) {
...
     public static void setLevel(String logger, Level level) {
-        // Setup log4j
-        Logger.getLogger(logger).setLevel(level);
+        loggerHardReferences.computeIfAbsent(logger, Logger::getLogger).setLevel(level);
+    }
 
-        // Setup java.util.logging
-        java.util.logging.Logger.getLogger(logger).setLevel(convertLog4jLevel(level));
+    private void log(Level level, Object msg) {
+        StackTraceElement caller = new Throwable().getStackTrace()[2];
```
sabieber, line 146: Ist das sicher dass da keine NPE / IndexOutOfBounds auftritt?

src/main/java/sirius/kernel/settings/ExtendedSettings.java
```diff
@@ -160,12 +168,13 @@ protected Extension getDefault(String type) {
...
         }
 
-        Collections.sort(list);
-        for (Extension ex : list) {
+        Collections.sort(extensions);
+
         Map<String, Extension> result = new LinkedHashMap<>();
+        for (Extension ex : extensions) {
             result.put(ex.getId(), ex);
         }
```
sabieber, line 176:
```suggestion
        for (Extension extension : extensions) {
            result.put(extension.getId(), extension);
        }
```

src/main/java/sirius/kernel/health/Log.java
andyHa: mostly yes - this is a private method, therefore it has to be called by another internal method. as this class hasn't a main method, this other method also has to have a caller.

one might be able to break this by forking a thread and passing in a curryed (curried?) method handle - but who would ever do this?

PR #268, opened 2020-06-09
Prevents logging a wrong message.

andyHa:
If we failed to auto-detect any parts and the
programmer also didn't specify any, theres is
no need to complain that "an empty list" was
specified - we rather have to report that no
parts could be discovered (which is done below).

### DIFF
--- a/src/main/java/sirius/kernel/timer/EndOfDayTask.java
+++ b/src/main/java/sirius/kernel/timer/EndOfDayTask.java
@@ -9,6 +9,7 @@
 package sirius.kernel.timer;
 
 import sirius.kernel.commons.Timeout;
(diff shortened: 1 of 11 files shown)


src/main/java/sirius/kernel/timer/EndOfDayTask.java
```diff
@@ -29,6 +30,7 @@
  * the next task. If a single task has the possibility of running for a long period (e.g. greater than one hour),
  * the task itself should protect the system (e.g. by limiting its runtime via {@link Timeout}.
  */
+@AutoRegister
```
sabieber, line 33: Das in nem Code Formatting Commit ist mutig ;)

PR #274, opened 2020-06-29
Log an exception that terminated the application silently

qw3ry:


src/main/java/sirius/kernel/Setup.java
```diff
@@ -132,7 +132,12 @@ public static void createAndStartEnvironment(ClassLoader loader) {
      * @param args the command line arguments (currently ignored)
      */
     public static void main(String[] args) {
-        Sirius.start(new Setup(Mode.DEV, Setup.class.getClassLoader()));
+        try {
+            Sirius.start(new Setup(Mode.DEV, Setup.class.getClassLoader()));
+        } catch (Exception e) {
+            Sirius.LOG.SEVERE("Unknown starup error: " + e.getLocalizedMessage());
```
sabieber, line 138:
```suggestion
            Sirius.LOG.SEVERE("Unknown startup error: " + e.getLocalizedMessage());
```

PR #281, opened 2020-07-13
Shows the actual expected format in the error message

meg1502:
When parsing LocalTimes previously, even when we expected '9:00 am' from an user, if he failed to do so, the error message showed we expect 09:00.


meg1502:
I'm not quite sure if the solution in this PR is better than just translating the error message with the matching format into the en locale

PR #286, opened 2020-07-31
Adds methods to hash a file / inputstream in Hasher.java

meg1502:


src/main/java/sirius/kernel/commons/Hasher.java
```diff
@@ -157,6 +162,40 @@ public Hasher hashBytes(byte[] data, int offset, int length) {
...
+        byte[] buffer = new byte[8192];
+
+        int readBytes;
+        do {
+            readBytes = inputStream.read(buffer);
+            if (readBytes > 0) {
+                this.digest.update(buffer, 0, readBytes);
+            }
+        } while (readBytes != -1);
+        inputStream.close();
```
andyHa, line 183: Shouldn‘t we let the caller close the stream? Zo encourage using try-with-resources

src/main/java/sirius/kernel/commons/Hasher.java
```diff
@@ -157,6 +162,40 @@ public Hasher hashBytes(byte[] data, int offset, int length) {
...
+
+    /**
+     * Reads the given file and appends its bytes to the data to be hashed.
+     *
+     * @param file the file to be read and put into the hash function
+     * @return the hasher itself for fluent method calls
+     * @throws IOException in case of an IO error while reading the file
+     */
+    public Hasher hashFile(File file) throws IOException {
+        hashStream(new FileInputStream(file));
```
andyHa, line 195: S.a. wrap in try block

PR #287, opened 2020-08-11
Returns optional Amounts

idlira:
Adds a helper method to return an Amount as optional, aka to asOptionalString and asOptionalInt.

### DIFF
--- a/src/main/java/sirius/kernel/commons/Value.java
+++ b/src/main/java/sirius/kernel/commons/Value.java
@@ -345,7 +345,7 @@ public class Value {
         if (Amount.NOTHING.equals(result)) {
             return Optional.empty();
         }
-        return Optional.ofNullable(getAmount());
+        return Optional.ofNullable(result);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


idlira:
> Does that really make sense? Value.of(null).getAmount() should return Amount.NOTHING so if i'm correct you'll never get an Optional.empty, only an Amount.NOTHING packaged into an Optional, and i don't know if thats really helpful

Value.of(null).getAmount() currently returns null. I do get an empty optional as of now.
If that should return Amount.NOTHING, is then a different story 😊

src/main/java/sirius/kernel/commons/Value.java
```diff
@@ -341,6 +341,10 @@ public Value replaceIfEmpty(@Nonnull Supplier<?> supplier) {
      */
     @Nonnull
     public Optional<Amount> asOptionalAmount() {
+        Amount result = getAmount();
+        if (Amount.NOTHING.equals(result)) {
+            return Optional.empty();
+        }
         return Optional.ofNullable(getAmount());
```
meg1502, line 348:
```suggestion
        return Optional.ofNullable(result);
```

PR #288, opened 2020-08-11
Defines alternative methods to Amount

idlira:
void ifPresent(consumer)
void ifEmpty(runnable)
Amount orElse(amount)
Amount orElseGet(supplier) -> making computeIfNull deprecated

### DIFF
--- a/src/main/java/sirius/kernel/commons/Amount.java
+++ b/src/main/java/sirius/kernel/commons/Amount.java
@@ -253,17 +253,16 @@ public class Amount implements Comparable<Amount> {
      *
      * @param supplier the supplier which is used to compute a value if there is no internal value
      * @return <tt>this</tt> if there is an internal value, the computed value of <tt>supplier</tt> otherwise
+     * @deprecated This method has been deprecated. Use <tt>orElseGet()</tt> instead.
      */
     @Nonnull
+    @Deprecated
     public Amount computeIfNull(Supplier<Amount> supplier) {
-        if (isEmpty()) {
-            return supplier.get();
-        }
-        return this;
+        return orElseGet(supplier);
     }
 
     /**
-     * Invokes the given consumer with the internal value is not empty.
+     * Invokes the given consumer if the internal value is not empty.
      *
      * @param consumer the consumer to execute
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/kernel/commons/Amount.java
```diff
@@ -261,6 +262,41 @@ public Amount computeIfNull(Supplier<Amount> supplier) {
...
     }
 
+    /**
+     * Invokes the given consumer with the internal value is not empty.
+     *
+     * @param consumer the consumer to execute
+     */
+    public void ifPresent(Consumer<Amount> consumer) {
+        if (isFilled()) {
+            consumer.accept(Amount.of(getAmount()));
```
meg1502, line 272: Can't we just use 'this' here?

src/main/java/sirius/kernel/commons/Amount.java
```diff
@@ -261,6 +262,41 @@ public Amount computeIfNull(Supplier<Amount> supplier) {
...
+        }
+    }
+
+    /**
+     * Returns the provided alternative {@link Amount} if the internal value is empty.
+     *
+     * @param amount the alternative Amount to return
+     * @return the original or alternative amount
+     */
+    public Amount orElseGet(Amount amount) {
```
meg1502, line 293: Rename to orElse or change Parameter to Supplier

idlira: Renamed to orElse.
orElseGet already exists in form of computeIfNull.
Can create a new method (to keep the naming standards) and mark the other as deprecated or just leave as is.

src/main/java/sirius/kernel/commons/Amount.java
```diff
@@ -261,6 +262,41 @@ public Amount computeIfNull(Supplier<Amount> supplier) {
         return this;
     }
 
+    /**
+     * Invokes the given consumer with the internal value is not empty.
```
bwiedmann, line 266:
```suggestion
     * Invokes the given consumer if the internal value is not empty.
```

PR #290, opened 2020-08-31
Cleanup Soap Client

qw3ry:
Remove superfluous overload, add convenience methods for handling namespaces

### DIFF
--- a/src/main/java/sirius/kernel/xml/StructuredOutput.java
+++ b/src/main/java/sirius/kernel/xml/StructuredOutput.java
@@ -34,17 +34,6 @@ public interface StructuredOutput {
     StructuredOutput beginResult(@Nonnull String name);
 
     /**
-     * Convenience method for {@link #beginResult(String)} prepending a namespace.
-     *
-     * @param namespace the namespace
-     * @param name      the unqualified name
-     * @return the output itself for fluent method calls
-     */
-    default StructuredOutput beginResult(@Nonnull String namespace, @Nonnull String name) {
-        return beginResult(namespace + ":" + name);
-    }
-
-    /**
      * Finishes (closes) the result
      */
     void endResult();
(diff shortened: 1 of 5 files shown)


src/main/java/sirius/kernel/xml/StructuredOutput.java
```diff
@@ -71,6 +86,18 @@
...
 
+    /**
+     * Convenience method for {@link #property(String, Object)} prepending a namespace.
+     *
+     * @param namespace the namespace
+     * @param name      the name of the property
+     * @param data      the value of the property
+     * @return the output itself for fluent method calls
+     */
+    default StructuredOutput property(@Nonnull String namespace, @Nonnull String name, @Nullable Object data) {
```
andyHa, line 97: shoudln't we move this into XMLStructuredOutput as JSON isn't namespace aware..

qw3ry: Good point +1

src/main/java/sirius/kernel/xml/StructuredOutput.java
```diff
@@ -38,14 +38,6 @@
      */
     void endResult();
 
-    /**
-     * Starts a new object with the given name.
-     *
-     * @param name the name of the element to start
-     * @return the output itself for fluent method calls
-     */
-    StructuredOutput beginObject(@Nonnull String name);
```
andyHa, line 47: sure about this? this would be a breaking change - deprecate otherwise?

andyHa (commented):
...

src/main/java/sirius/kernel/xml/StructuredOutput.java
qw3ry: It has an overload `StructuredOutput beginObject(@Nonnull String name, Attribute... attributes)` which will be called instead. A library user might need to recompile, but the code will still be compatible

andyHa (approved):

qw3ry:
"Breaking" because two methods became one, and cannot both be overridden. I will adjust the corresponding code in -web once this PR is merged

bwiedmann (approved):

qw3ry:
https://github.com/scireum/sirius-web/pull/747 required changes in web

meg1502 (approved):

andyHa (approved):

sabieber (approved):

mkeckmkeck (approved):

PR #292, opened 2020-09-08
Allows to differentiate between timings...

fhaScireum:
...which should be shown to all users and which should be shown to administrators only

- Fixes: SIRI-250

### DIFF
--- a/src/main/java/sirius/kernel/async/BasicTaskContextAdapter.java
+++ b/src/main/java/sirius/kernel/async/BasicTaskContextAdapter.java
@@ -57,11 +57,6 @@ public class BasicTaskContextAdapter implements TaskContextAdapter {
     }
 
     @Override
-    public void addTiming(String counter, long millis) {
-        // Ignored by the default implementation.
-    }
-
-    @Override
     public void addTiming(String counter, long millis, boolean adminOnly) {
         // Ignored by the default implementation.
     }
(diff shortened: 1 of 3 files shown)


qw3ry (approved):

src/main/java/sirius/kernel/async/BasicTaskContextAdapter.java
```diff
@@ -61,6 +61,11 @@ public void addTiming(String counter, long millis) {
         // Ignored by the default implementation.
```
andyHa, line 61: provide a default impl here

sabieber (approved):

src/main/java/sirius/kernel/async/BasicTaskContextAdapter.java
andyHa: or rather kill, das this method is never invoked?

src/main/java/sirius/kernel/async/TaskContext.java
```diff
@@ -167,6 +167,22 @@ public void addTiming(String counter, long millis) {
         adapter.addTiming(counter, millis);
```
andyHa, line 167: call  adapter.addTiming(counter, millis, false);

src/main/java/sirius/kernel/async/TaskContextAdapter.java
```diff
@@ -61,6 +61,15 @@
      */
     void addTiming(String counter, long millis);
```
andyHa, line 62: kill?

andyHa (changes requested):
...

idlira (approved):

andyHa (approved):

mkeckmkeck (approved):

PR #294, opened 2020-09-14
Added question mark check to URLBuilder

marvinwoop:
If adding a baseURL oder host containing a question mark the question mark monoflop will be toggled for following parameters to be added correctly with an ampersand.

### DIFF
--- a/src/main/java/sirius/kernel/commons/URLBuilder.java
+++ b/src/main/java/sirius/kernel/commons/URLBuilder.java
@@ -63,12 +63,13 @@ public class URLBuilder {
     /**
      * Adds a path part to the url.
      * <p>
-     * Once the first parameter has been added, the path can no longer be modified. If the path itself contains
-     * parameters the questionMark {@link Monoflop} will be toggled.
+     * Once the first parameter has been added, the path can no longer be modified. Also the part itself can (but
+     * shouldn't) contain parameters, usually led by a initial question mark.
      *
      * @param uriPartsToAdd the uri part to add. This should not contain a leading '/' as it is added automatically. If
      *                      an array (vararg) is given, all components are appended to the internal {@link
-     *                      StringBuilder} without any additional characters.
+     *                      StringBuilder} without any additional characters. If this contains a '?' to add parameters,
+     *                      no more parts can be added.
      * @return the builder itself for fluent method calls
      */
     public URLBuilder addPart(@Nonnull String... uriPartsToAdd) {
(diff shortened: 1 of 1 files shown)


idlira (approved):

src/main/java/sirius/kernel/commons/URLBuilder.java
```diff
@@ -63,7 +63,8 @@ public URLBuilder(@Nonnull String protocol, @Nonnull String host) {
     /**
      * Adds a path part to the url.
      * <p>
-     * Once the first parameter has been added, the path can no longer be modified.
+     * Once the first parameter has been added, the path can no longer be modified. If the path itself contains
+     * parameters the questionMark {@link Monoflop} will be toggled.
```
idlira, line 67: Not really sure that we need to explain this logic in the JavaDoc, especially making reference to an internal private property.

mkeckmkeck: I agree, we should not put internal behaviour into javadoc. Maybe add a comment to the parameter instead, saying that it can contain query parameter if they are properly started by '?'?

mkeckmkeck (approved):

andyHa (approved):

src/main/java/sirius/kernel/commons/URLBuilder.java
marvinwoop: I changed the JavaDoc a little to meet your expectations.

src/main/java/sirius/kernel/commons/URLBuilder.java
```diff
@@ -63,12 +63,13 @@ public URLBuilder(@Nonnull String protocol, @Nonnull String host) {
     /**
      * Adds a path part to the url.
      * <p>
-     * Once the first parameter has been added, the path can no longer be modified. If the path itself contains
-     * parameters the questionMark {@link Monoflop} will be toggled.
+     * Once the first parameter has been added, the path can no longer be modified. Also the part itself can (but
+     * shouldn't) contain parameters, usually led by a initial question mark.
```
idlira, line 67:
```suggestion
     * shouldn't) contain parameters, usually led by an initial question mark.
```

qw3ry (approved):

mkeckmkeck (approved):

src/main/java/sirius/kernel/commons/URLBuilder.java
marvinwoop: Thanks, corrected it.

PR #297, opened 2020-11-10
Fixes erroneous report of background loop overtime

qw3ry:
Ticket SIRI-278

The fix is not super pretty, I'd be glad if you had suggestions

### DIFF
--- a/src/main/java/sirius/kernel/async/BackgroundLoop.java
+++ b/src/main/java/sirius/kernel/async/BackgroundLoop.java
@@ -58,6 +58,7 @@ public abstract class BackgroundLoop {
     private volatile boolean enabled = true;
     private long lastExecutionAttempt;
     private String executionInfo = "-";
+    private boolean executing = false;
 
     /**
      * Returns the name of the loop.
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/kernel/async/BackgroundLoop.java
```diff
@@ -196,6 +199,15 @@ public String getExecutionInfo() {
         return executionInfo;
     }
 
+    /**
+     * whether the background loop is currently running
```
andyHa, line 203:
```suggestion
     * Determines whether the background loop is currently running.
```

src/main/java/sirius/kernel/async/BackgroundLoop.java
```diff
@@ -196,6 +199,15 @@ public String getExecutionInfo() {
...
+     *
+     * @return true if the background loop is running
```
andyHa, line 205:
```suggestion
     * @return <tt>true</tt> if the background loop is running, <tt>false</tt> otherwise
```

meg1502 (approved):

idlira (approved):

sabieber (approved):

marvinwoop (approved):

src/main/java/sirius/kernel/async/BackgroundLoop.java
```diff
@@ -58,6 +58,7 @@
     private volatile boolean enabled = true;
     private long lastExecutionAttempt;
     private String executionInfo = "-";
+    private boolean executing = false;
```
andyHa, line 61: volatile or AtomicBoolean?

PR #301, opened 2020-11-27
Fix writing to System.err

mkeckmkeck:
- Former implementation was first instantiating a ConsoleHandler with err and then setting output on the instance to out. That caused a close of the previously set err.
- Adapt constructor call and extend StreamHandler to prevent this

Fixes: SIRI-285

### DIFF
--- a/src/main/java/sirius/kernel/Setup.java
+++ b/src/main/java/sirius/kernel/Setup.java
@@ -244,7 +244,6 @@ public class Setup {
         }
 
         StreamHandler consoleHandler = new StdOutHandler();
-        consoleHandler.setFormatter(new SaneFormatter());
         consoleHandler.setLevel(Level.ALL);
 
         rootLogger.addHandler(consoleHandler);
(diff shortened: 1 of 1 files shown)


marvinwoop (approved):

idlira (approved):

andyHa (approved):

sabieber (approved):

src/main/java/sirius/kernel/Setup.java
```diff
@@ -243,18 +243,24 @@ protected void setupLogging() {
             rootLogger.removeHandler(handler);
         }
 
-        StdOutHandler consoleHandler = new StdOutHandler();
+        StreamHandler consoleHandler = new StdOutHandler();
         consoleHandler.setFormatter(new SaneFormatter());
```
jmuscireum, line 247: can be removed

jmuscireum (changes requested):

jmuscireum (approved):

PR #308, opened 2021-01-29
Provides an option to only read a node itself on xml parsing

mkeckmkeck:
- defaults to old behaviour of reading the node with all content

### DIFF
--- a/src/main/java/sirius/kernel/xml/NoContentNodeHandler.java
+++ b/src/main/java/sirius/kernel/xml/NoContentNodeHandler.java
@@ -1,0 +1,21 @@
+/*
+ * Made with all the love in the world
+ * by scireum in Remshalden, Germany
+ *
+ * Copyright by scireum GmbH
+ * http://www.scireum.de - info@scireum.de
+ */
+
+package sirius.kernel.xml;
+
+/**
+ * Can be used to handle XML nodes without having the parser to read out the full content of the node, but only
+ * the node with attributes. See {@link NodeHandler#ignoreContent()} for an example.
+ */
+public abstract class NoContentNodeHandler implements NodeHandler {
+
+    @Override
+    public boolean ignoreContent() {
+        return true;
+    }
+}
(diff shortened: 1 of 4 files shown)


src/main/java/sirius/kernel/xml/NoContentNodeHandler.java
```diff
@@ -0,0 +1,21 @@
...
+ * http://www.scireum.de - info@scireum.de
+ */
+
+package sirius.kernel.xml;
+
+/**
+ * Can be used to handle XML nodes without having the parser to read out the full content of the node, but only
+ * the node with attributes. See {@link NodeHandler#ignoreContent()} for an example.
+ */
+public abstract class NoContentNodeHandler implements NodeHandler {
```
idlira, line 15: I would then rename to NoChildrenNodeHandler, so its clear that the Node contains all attributes and text value.

src/main/java/sirius/kernel/xml/NodeHandler.java
```diff
@@ -16,9 +16,32 @@
...
-     * Invoked once a complete subtree was parsed
+     * Invoked once a node was parsed. Depending on {@link #ignoreContent()} this will be the complete subtree or only
+     * the node itself.
      *
      * @param node the root node of the subtree parsed by the SAX parser
      */
     void process(StructuredNode node);
+
+    /**
+     * Determines if the content of an node shall be handled. It can be used for root nodes of large files. There we do
```
sabieber, line 27:
```suggestion
     * Determines if the content of a node shall be handled. It can be used for root nodes of large files. There we do
```

idlira (changes requested):
name suggestion

sabieber (changes requested):

andyHa (approved):

src/main/java/sirius/kernel/xml/NoContentNodeHandler.java
mkeckmkeck: Actually it does not, see latest commit for clarification inside javadoc example. In case we need a more sophisticated handling with xml mixed mode nodes, I suggest we add that at a later point.

idlira (approved):

sabieber (approved):

marvinwoop (approved):

qw3ry (approved):

PR #317, opened 2021-03-03
Stops auto transforming GET requests into POST Requests and adds some methods to Outcall

meg1502:
BREAKING: No longer provides access to outputstream for get requests

Previously this would work, as the request was transformed into POST from GET if you called getOutput on a GET connection. As we have an explicit method to transform a connection into a POST connection, we excpect this method to be called instead of relying on this auto transform (which is marked as backwards compatibility in the java API).

So to be save that this does not break things, you should search the accessors for Outcall.getOutput in your project and make sure that markAsPostRequest() is called before

### DIFF
--- a/src/main/java/sirius/kernel/xml/Outcall.java
+++ b/src/main/java/sirius/kernel/xml/Outcall.java
@@ -41,6 +41,7 @@ import java.security.SecureRandom;
 import java.time.Duration;
 import java.util.Base64;
 import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 import java.util.regex.Matcher;
@@ -60,8 +61,11 @@ public class Outcall {
     private static final String REQUEST_METHOD_POST = "POST";
     private static final String REQUEST_METHOD_HEAD = "HEAD";
     private static final String HEADER_CONTENT_TYPE = "Content-Type";
+    private static final String HEADER_CONTENT_DISPOSITION = "Content-Disposition";
     private static final String CONTENT_TYPE_FORM_URLENCODED = "application/x-www-form-urlencoded; charset=utf-8";
     private static final Pattern CHARSET_PATTERN = Pattern.compile("(?i)\\bcharset=\\s*\"?([^\\s;\"]*)");
+    private static final Pattern CONTENT_DISPOSITION_FILENAME_PATTERN =
+            Pattern.compile("(attachment|inline);\\s*filename\\s*=\\s*\"([^\"]*)\"");
     private static final X509TrustManager TRUST_SELF_SIGNED_CERTS = new TrustingSelfSignedTrustManager();
 
     /**
@@ -239,7 +243,7 @@ public class Outcall {
      * Note that you need to call {@link #markAsPostRequest()} before calling this method.
      *
      * @return the stream of data sent to the call / url
-     * @throws IOException in case of any IO error
+     * @throws IOException                in case of any IO error
      * @throws java.net.ProtocolException if the method doesn't support output
      */
     public OutputStream getOutput() throws IOException {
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/kernel/xml/Outcall.java
```diff
@@ -337,6 +377,17 @@ public String getHeaderField(String name) {
...
     }
 
+    /**
+     * Returns the response header with the given name.
+     *
+     * @param name        the name of the header to fetch
+     * @param defaultDate a default value to return if the header field is missing or malformed.
+     * @return the value of the given header in the response or the given <tt>defaultDate</tt>
+     */
+    public long getHeaderFieldDate(String name, long defaultDate) {
```
andyHa, line 387: maybe getHeaderAsTimestamp or the like? as we do not return a date but (presumably an UTC timestamp)

src/main/java/sirius/kernel/xml/Outcall.java
```diff
@@ -246,6 +273,19 @@ public Outcall setRequestProperty(String name, String value) {
...
+    /**
+     * Sets the value of the {@code ifModifiedSince} header of this connection.
+     * <p>
+     * The fetching of the object is skipped unless the object has been modified more recently than a certain time.
+     * If the object will not be returned because of this, the response code will be <tt>304</tt>.
+     *
+     * @param ifModifiedSince a date since when the object should be modified
+     * @throws IllegalStateException if already connected
+     */
+    public void setIfModifiedSince(long ifModifiedSince) {
```
andyHa, line 285: add a helper for LocalDateTime ? or at least Instant?

src/main/java/sirius/kernel/xml/Outcall.java
andyHa: oooor return an Optional<LocalDateTime> or the like

sabieber (approved):

src/main/java/sirius/kernel/xml/Outcall.java
```diff
@@ -165,10 +166,23 @@ public Outcall postData(Context params, Charset charset) throws IOException {
...
         return this;
     }
 
+    /**
+     * Marks the request as HEAD request, only requesting headers.
+     *
+     * @return the outcall itself for fluent method calls
+     * @throws IOException if the method cannot be reset or if the requested method isn't valid for HTTP.
+     */
+    public Outcall onlyRequestHeaders() throws IOException {
```
andyHa, line 180: markAsHeadRequest (?)

src/main/java/sirius/kernel/xml/Outcall.java
```diff
@@ -165,10 +166,23 @@ public Outcall postData(Context params, Charset charset) throws IOException {
...
      */
     public Outcall markAsPostRequest() throws IOException {
+        connection.setDoOutput(true);
         connection.setRequestMethod(REQUEST_METHOD_POST);
         return this;
     }
 
+    /**
+     * Marks the request as HEAD request, only requesting headers.
+     *
```
andyHa, line 176:
```suggestion
     * <p>
     * Note that neither {@link #getInput()} nor {@link #getOutput()} can be invoked on this call.
```

idlira (approved):
Generally I find good to describe the impact and actions to be taken in the PR's description, making it easier to be ported into the final products.

src/main/java/sirius/kernel/xml/Outcall.java
```diff
@@ -58,9 +64,14 @@
     private static final int DEFAULT_READ_TIMEOUT = (int) TimeUnit.MILLISECONDS.convert(5, TimeUnit.MINUTES);
 
     private static final String REQUEST_METHOD_POST = "POST";
+    private static final String REQUEST_METHOD_HEAD = "HEAD";
     private static final String HEADER_CONTENT_TYPE = "Content-Type";
+    private static final String HEADER_CONTENT_DISPOSITION = "Content-Disposition";
```
andyHa, line 69: either content-disposition or If-Modified-Since below....^^

src/main/java/sirius/kernel/xml/Outcall.java
```diff
@@ -337,6 +391,41 @@ public String getHeaderField(String name) {
...
+     * @return an Optional containing the file name given by the header, or Optional.empty if no file name is given
+     */
+    public Optional<String> parseFileNameFromContentDisposition() {
+        try {
+            Matcher matcher =
+                    CONTENT_DISPOSITION_FILENAME_PATTERN.matcher(connection.getHeaderField(HEADER_CONTENT_DISPOSITION));
+            if (matcher.find()) {
+                return Optional.ofNullable(matcher.group(2));
+            }
+        } catch (IllegalStateException | IndexOutOfBoundsException e) {
```
andyHa, line 423: why would these occur? the pattern is guaranteed to have 3 groups and the state is consistent if find returns true (?)

sabieber (approved):

bwiedmann (approved):

sabieber (approved):

idlira (approved):

bwiedmann (approved):

cschierle (approved):

mkeckmkeck (approved):

PR #319, opened 2021-03-08
Stops NPE when trying to parse a non-existent content-disposition header

meg1502:


src/main/java/sirius/kernel/xml/Outcall.java
```diff
@@ -414,10 +414,12 @@ public String getHeaderField(String name) {
      * @return an Optional containing the file name given by the header, or Optional.empty if no file name is given
      */
     public Optional<String> parseFileNameFromContentDisposition() {
-        Matcher matcher =
-                CONTENT_DISPOSITION_FILENAME_PATTERN.matcher(connection.getHeaderField(HEADER_CONTENT_DISPOSITION));
-        if (matcher.find()) {
-            return Optional.ofNullable(matcher.group(2));
+        String contentDisposition = connection.getHeaderField(HEADER_CONTENT_DISPOSITION);
+        if (Strings.isFilled(contentDisposition)) {
```
idlira, line 418: invert and return Optional.empty() if Strings.isEmpty?

sabieber (approved):

fhaScireum (approved):

src/main/java/sirius/kernel/xml/Outcall.java
qw3ry: I like it more this way, because you don't need an additional `return Optional.empty()`, but I guess this comes down to personal preference

qw3ry (approved):

idlira (approved):

PR #322, opened 2021-03-24
Compatibility for filename without quotes in Content-Disposition header

meg1502:
Fixes:OX-6612

### DIFF
--- a/src/main/java/sirius/kernel/xml/Outcall.java
+++ b/src/main/java/sirius/kernel/xml/Outcall.java
@@ -71,7 +71,7 @@ public class Outcall {
     private static final String CONTENT_TYPE_FORM_URLENCODED = "application/x-www-form-urlencoded; charset=utf-8";
     private static final Pattern CHARSET_PATTERN = Pattern.compile("(?i)\\bcharset=\\s*\"?([^\\s;\"]*)");
     private static final Pattern CONTENT_DISPOSITION_FILENAME_PATTERN =
-            Pattern.compile("(attachment|inline);\\s*filename\\s*=\\s*(?<unquoted>\"(?<quoted>[^\"]*)\"|[\\w.-]+)");
+            Pattern.compile("(attachment|inline|form-data);.*filename\\s*=\\s*(?<unquoted>\"(?<quoted>[^\"]*)\"|[\\w.-]+)");
     private static final X509TrustManager TRUST_SELF_SIGNED_CERTS = new TrustingSelfSignedTrustManager();
 
     /**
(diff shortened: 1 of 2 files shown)


sabieber (approved):

fhaScireum (approved):

src/main/java/sirius/kernel/xml/Outcall.java
```diff
@@ -71,7 +71,7 @@
     private static final String CONTENT_TYPE_FORM_URLENCODED = "application/x-www-form-urlencoded; charset=utf-8";
     private static final Pattern CHARSET_PATTERN = Pattern.compile("(?i)\\bcharset=\\s*\"?([^\\s;\"]*)");
     private static final Pattern CONTENT_DISPOSITION_FILENAME_PATTERN =
-            Pattern.compile("(attachment|inline);\\s*filename\\s*=\\s*\"([^\"]*)\"");
+            Pattern.compile("(attachment|inline);\\s*filename\\s*=\\s*\"?([^\"]*)\"?");
```
qw3ry, line 74: Sicher, dass das so funktioniert?
- Der Filename kann damit das " auch nur vorne ODER hinten haben
- Wenn der Filename ohne " endet, geht der Match trotzdem bis zum nächsten " weiter. Also `filename = test.pdf, charset = "utf8"` würde `test.pdf, charset =` matchen

qw3ry: Eventuell besser `\"([^\"]*)\"|\w+`, das _sollte_ einige Probleme vermeiden

sabieber (approved):

qw3ry (approved):

src/main/java/sirius/kernel/xml/Outcall.java
```diff
@@ -71,7 +71,7 @@
     private static final String CONTENT_TYPE_FORM_URLENCODED = "application/x-www-form-urlencoded; charset=utf-8";
     private static final Pattern CHARSET_PATTERN = Pattern.compile("(?i)\\bcharset=\\s*\"?([^\\s;\"]*)");
     private static final Pattern CONTENT_DISPOSITION_FILENAME_PATTERN =
-            Pattern.compile("(attachment|inline);\\s*filename\\s*=\\s*\"([^\"]*)\"");
+            Pattern.compile("(attachment|inline|form-data);.*filename\\s*=\\s*(?<unquoted>\"(?<quoted>[^\"]*)\"|[\\w.-]+)");
```
idlira, line 74: Is it wise to read it from a config setting?
It looks to me that very soon we will find yet another website which delivers other stuff.

meg1502: this should be a catch all approach, if we find stuff this doesn't match we need to add it to the test and expand the pattern

idlira (approved):

src/main/java/sirius/kernel/xml/Outcall.java
andyHa: A config value can also only changed via a patch and we probably do not want this to differ from product to product

andyHa (approved):

PR #328, opened 2021-04-15
Output namespaced xml property only if value is not null

mkeckmkeck:


src/main/java/sirius/kernel/xml/XMLStructuredOutput.java
```diff
@@ -297,6 +297,28 @@ public StructuredOutput namespacedProperty(@Nonnull String namespace,
         return property(namespace + ":" + name, data, attributes);
     }
 
+    /**
+     * Convenience method for {@link #property(String, Object, Attribute...)} prepending a namespace.
```
sabieber, line 301:
```suggestion
     * Provides a convenient way for {@link #property(String, Object, Attribute...)} with a prepended namespace.
```

sabieber (approved):

andyHa (approved):

idlira (approved):

src/main/java/sirius/kernel/xml/XMLStructuredOutput.java
mkeckmkeck: done for all copy&paste convenient texts

sabieber (approved):

ymo-sci (approved):

PR #338, opened 2021-06-24
Deprecates methods related to language validation in NLS.

cschierle:
These will be replaced by an appropriate solution in UserManager of sirius-web (see https://github.com/scireum/sirius-web/pull/864)

Fixes: SIRI-368

### DIFF
--- a/src/main/java/sirius/kernel/async/CallContext.java
+++ b/src/main/java/sirius/kernel/async/CallContext.java
@@ -9,7 +9,6 @@
 package sirius.kernel.async;
 
 import sirius.kernel.Sirius;
-import sirius.kernel.commons.Explain;
 import sirius.kernel.commons.Strings;
 import sirius.kernel.commons.Tuple;
 import sirius.kernel.commons.Value;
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/kernel/async/CallContext.java
```diff
@@ -359,6 +360,8 @@ public void removeFromMDC(String key) {
      *
      * @return a two-letter language code used for the current thread.
      */
+    @SuppressWarnings("deprecation")
+    @Explain("The replacement for NLS.getDefaultLanguage() is only available in sirius-web.")
```
andyHa, line 364: I'd say this should remain as sirius-web will provide a lazyLanguageInstaller (?)

src/main/java/sirius/kernel/nls/NLS.java
```diff
@@ -131,6 +131,7 @@ public static String getCurrentLang() {
      * @return the language code of the default language
      */
     @Nonnull
+    @Deprecated
     public static String getDefaultLanguage() {
```
andyHa, line 135: shouldn't this remain?

andyHa (changes requested):

src/main/java/sirius/kernel/nls/NLS.java
cschierle: Hmm... Depends on what the `defaultLanguage` actually, or rather if it is defined by the NLS or the scope settings. I'll look into this during further development

andyHa (approved):

sabieber (approved):

idlira (approved):

mkeckmkeck (approved):

mkeckmkeck:
Potetially breaking when deprecation usage not allowed

PR #344, opened 2021-07-29
Allows to disconnect an Outcall before completion.

meg1502:
Fixes: OX-7063

This may be useful if the connection is only used to read the headers


sabieber (approved):
PR Improvements made:
- Added labels
- Improved title

andyHa (approved):

PR #345, opened 2021-08-02
Set DOCKER_HOST environment for docker-compose

mkeckmkeck:


src/main/resources/component-kernel.conf
```diff
@@ -319,10 +319,19 @@ docker {
...
+    # 'DOCKER_HOST' will be prepended with TCP protocol and appended with #docker.hostPort.
+    # E.g.: DOCKER_HOST = "tcp://<IP>:2375"
+    # If left empty, docker-compose will default to local docker daemon via socket
+    # (like unix:///var/run/docker.sock Note: that is no valid configuration as IP is missing).
     hostIp = ""
     hostIp = ${?SIRIUS_DOCKER_HOST}
 
+    # Used in combination with #docker.hostIp to determine the 'DOCKER_HOST' TCP address.
+    hostPort = "2375"
+    hostPort = ${?SIRIUS_DOCKER_HOST}
```
meg1502, line 333: is this the right key? it's the same as for hostIp

meg1502 (changes requested):

sabieber (approved):

src/main/resources/component-kernel.conf
sabieber: Probably not

mkeckmkeck: thx

meg1502 (approved):

ymo-sci (approved):

PR #350, opened 2021-08-23
Adds support for protocol changing HTTP redirects.

andyHa:
HttpUrlConnection does not follow redirects if the protocol is changed.
We fix this, as redirects (from HTTP to HTTPS) are quite common.

The implementation is heavily inspired by
https://mkyong.com/java/java-httpurlconnection-follow-redirect-example/

Fixes: SIRI-412


meg1502 (changes requested):
We probably also need to call connect in the getHeaderField methods?

meg1502 (approved):

sabieber (approved):

qw3ry (approved):

idlira (approved):

PR #361, opened 2021-10-05
Prevents that one accidentally adds a Value or Optional to a context.

andyHa:
In the past, this was a bug in 100% of all cases, as consumers of the
Context normally expect plain types and not ones wrapped in a Value
or Optional. Also there is no actual benefit in doing so - however, there
is no real support in the Java language to prevent this.

Therefore we add these checks to at least "fail early and notably".

### DIFF
--- a/src/main/java/sirius/kernel/commons/Context.java
+++ b/src/main/java/sirius/kernel/commons/Context.java
@@ -12,6 +12,7 @@ import javax.annotation.Nonnull;
 import javax.script.ScriptContext;
 import java.util.Collection;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.TreeMap;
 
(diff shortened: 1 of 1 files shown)


jakobvogel (approved):

fhaScireum (approved):

src/main/java/sirius/kernel/commons/Context.java
```diff
@@ -172,6 +173,13 @@ public static Context create() {
      * @return <tt>this</tt> to permit fluent method calls
```
qw3ry, line 173: maybe add a sentence to the docs?

qw3ry (approved):
If there is a use case for Optionals/Values in the context, then we could also add a second setter method `forceSet` or something that bypasses the check. But I cannot think of sane use cases anyways

sabieber (approved):

mkeckmkeck (approved):

idlira (approved):

meg1502 (approved):

PR #374, opened 2021-12-06
Adds a helper that extracts a LocalDate from a Value

cschierle:


sabieber (approved):

src/main/java/sirius/kernel/nls/NLS.java
```diff
@@ -973,16 +973,63 @@ private static String formatSpokenRecentDateWithTime(LocalDateTime givenDateTime
...
+     *
+     * @param value                  a {@link Value} object containing the date to be parsed
+     * @param formatter              the formatter to use for parsing
+     * @param fallbackLanguageFormat the language providing a fallback date format
+     * @return the date or <tt>null</tt> if there is no source value to parse
+     * @throws DateTimeParseException if the given value could not be parsed as a date
+     */
+    @Nullable
+    public static LocalDate extractDate(@Nonnull Value value,
+                                        DateTimeFormatter formatter,
```
idlira: Is it of interest to have a version without formatter?
And should it replicate the logic from here? https://github.com/scireum/scireum-core/blob/e3e2152855afc83a0811584adb4bff5bb256ebeb/src/main/java/scireum/core/data/CustomPropertyType.java#L583

idlira (changes requested):

src/main/java/sirius/kernel/nls/NLS.java
cschierle: Since the method is now moved to `Value` the version without formatter would be `value.asLocalDate()`.

And the linked logic does not apply for this use case. During an import, a date String using a certain format is expected. It is not guaranteed that a value is given in the expected format due to user error.
So we are always in the `expectUserInput`-path of your example, but want to provide a reasonable amount of fallback options, to not reject a date if it was not given exactly in the expected format.

idlira: OK... but than we could include more parsing to the fallback options.
NLS.parseMachineString includes the ISO case: (YYYY-MM-DD)
and NLS.parseUserString uses the current language, which could in theory replace one of your fallbacks?

sabieber: I think we should do that when needed

idlira (approved):

sabieber (approved):

src/main/java/sirius/kernel/commons/Value.java
```diff
@@ -1298,6 +1300,46 @@ public LocalDateTime asLocalDateTimeOfEpochSeconds(LocalDateTime defaultValue) {
         return LocalDateTime.ofInstant(temporal, ZoneId.systemDefault());
     }
 
+    /**
+     * Parses a date using the given formatter. Tries parsing with the format for the given fallback language in case
```
andyHa, line 1304: this method looks overly complex.

Maybe we add two versions:
public Value tryParseDate(DateTimeFormatter formatter) { ... } and
public Value tryParseDate(String languageCode) { return tryParseDate(NLS.getDateFormat(languageCode)); }

The first method always checks for null or if the wrapped value is already a local date and then just returns self.

This way, one could replicate the logic by calling value.tryParseDate(FORMAT).tryParseDate("en").tryParseDate(NLS.getCurrentLanguage()).asOptional().orElseThrow("Invalid Date Format")

cschierle: Great suggestion 👍  This simplifies the code here a lot.

With these changes, a parsing retry-chain now looks like this:
```
value.tryParseLocalDate(DATE_FORMATTER)
     .tryParseLocalDate("fr")
     .tryParseLocalDate(NLS.getCurrentLang())
     .tryParseLocalDate(DateTimeFormatter.ISO_LOCAL_DATE)
     .asOptional(LocalDate.class)
     .orElseThrow(() -> Exceptions.createHandled()
                                  .withNLSKey("invalidDateFormat")
                                  .set("expectedFormat", DATE_FORMAT)
                                  .handle());

**Review by idlira — APPROVED** [review]

**Review by sabieber — APPROVED** [review]

File: src/main/java/sirius/kernel/commons/Value.java [hunk 3]
```diff
@@ -1301,45 +1301,36 @@ public LocalDateTime asLocalDateTimeOfEpochSeconds(LocalDateTime defaultValue) {
...
-                    return LocalDate.parse(getString(), NLS.getDateFormat(NLS.getCurrentLang()));
-                } catch (DateTimeParseException ex) {
-                    try {
-                        return LocalDate.parse(getString());
-                    } catch (DateTimeParseException exc) {
-                        throw exception;
-                    }
-                }
-            }
+            return this;
```
andyHa, line 1320: +Exceptions.ignore(exception);

andyHa (approved):

PR #375, opened 2021-12-09
Outcall timeout extension

marvinwoop:
1. The outcall timeouts, which take effect by default, are now configurable.
2. Additionally there are now so called interactive timeouts, which are meant to be applied on time-critical calls. These are calls which a user is actively waiting for. A call can simply be set interactive by calling `setInteractive()` and it will use the interactive timeout defaults.

### DIFF
--- a/src/main/java/sirius/kernel/xml/Outcall.java
+++ b/src/main/java/sirius/kernel/xml/Outcall.java
@@ -98,7 +98,7 @@ public class Outcall {
     @ConfigValue("http.outcall.timeouts.default.readTimeout")
     private static Duration defaultReadTimeout;
 
-    @ConfigValue("http.outcall.timeouts.connectTimeoutBlacklistDuration")
+    @ConfigValue("http.outcall.connectTimeoutBlacklistDuration")
     private static Duration connectTimeoutBlacklistDuration;
 
     private HttpClient client;
--- a/src/main/java/sirius/kernel/xml/Outcall.java
+++ b/src/main/java/sirius/kernel/xml/Outcall.java
@@ -344,16 +344,8 @@ public class Outcall {
     public Outcall withConfiguredTimeout(@Nonnull String configKey) {
         Extension extension = Sirius.getSettings().getExtension("http.outcall.timeouts", configKey);
 
-        if (extension != null) {
-            setConnectTimeout((int) extension.getConfig().getDuration("connectTimeout").toMillis());
-            setReadTimeout((int) extension.getConfig().getDuration("readTimeout").toMillis());
-        } else {
-            throw Exceptions.handle()
-                            .to(Log.SYSTEM)
-                            .withSystemErrorMessage("For the following config string no timeout configuration and "
-                                                    + "default block could be found: %s", configKey)
-                            .handle();
-        }
+        setConnectTimeout((int) extension.getConfig().getDuration("connectTimeout").toMillis());
+        setReadTimeout((int) extension.getConfig().getDuration("readTimeout").toMillis());
 
         return this;
     }
--- a/src/main/java/sirius/kernel/xml/Outcall.java
+++ b/src/main/java/sirius/kernel/xml/Outcall.java
@@ -21,6 +21,7 @@ import sirius.kernel.health.Exceptions;
 import sirius.kernel.health.Log;
 import sirius.kernel.health.Microtiming;
 import sirius.kernel.nls.NLS;
+import sirius.kernel.settings.Extension;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
@@ -341,19 +342,16 @@ public class Outcall {
      * @return this for fluent method calls
      */
     public Outcall withConfiguredTimeout(@Nonnull String configKey) {
-        Duration connectTimeout =
-                Sirius.getSettings().getDuration(Strings.apply("http.outcall.timeouts.%s.connectTimeout", configKey));
-        Duration readTimeout =
-                Sirius.getSettings().getDuration(Strings.apply("http.outcall.timeouts.%s.readTimeout", configKey));
+        Extension extension = Sirius.getSettings().getExtension("http.outcall.timeouts", configKey);
 
-        if (!connectTimeout.isZero() && !readTimeout.isZero()) {
-            setConnectTimeout((int) connectTimeout.toMillis());
-            setReadTimeout((int) readTimeout.toMillis());
+        if (extension != null) {
+            setConnectTimeout((int) extension.getConfig().getDuration("connectTimeout").toMillis());
+            setReadTimeout((int) extension.getConfig().getDuration("readTimeout").toMillis());
         } else {
             throw Exceptions.handle()
                             .to(Log.SYSTEM)
-                            .withSystemErrorMessage("A timeout configuration could not be found for config string: %s",
-                                                    configKey)
+                            .withSystemErrorMessage("For the following config string no timeout configuration and "
+                                                    + "default block could be found: %s", configKey)
                             .handle();
         }
 
--- a/src/main/java/sirius/kernel/xml/Outcall.java
+++ b/src/main/java/sirius/kernel/xml/Outcall.java
@@ -94,6 +94,13 @@ public class Outcall {
 
     @ConfigValue("http.outcall.defaultReadTimeout")
     private static Duration defaultReadTimeout;
+
+    @ConfigValue("http.outcall.interactive.connectTimeout")
+    private static Duration interactiveConnectTimeout;
+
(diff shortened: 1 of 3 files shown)


sabieber (approved):

src/main/java/sirius/kernel/xml/Outcall.java
```diff
@@ -92,6 +89,18 @@
      */
     private static final int TIMEOUT_BLACKLIST_HIGH_WATERMARK = 100;
 
+    @ConfigValue("http.outcall.defaultConnectTimeout")
```
andyHa, line 92: Isn' this a bit of an overkill here? We can already overwrite the default timeouts if needed...

src/main/java/sirius/kernel/xml/Outcall.java
```diff
@@ -325,6 +334,14 @@ public void setReadTimeout(int timeoutMillis) {
         modifyRequest().timeout(Duration.ofMillis(timeoutMillis));
     }
 
+    /**
+     * Sets the connect and read timeout to a interactively reasonable duration defined in http.outcall.interactive.*
+     */
+    public void setInteractive() {
```
andyHa, line 340: Never call a method setXXX unless it is a setter. Also, why not simply provide a proper timeout here instead of calling a magic method....

src/main/resources/component-kernel.conf
```diff
@@ -117,9 +117,25 @@ http.outcall {
...
+    # The default outcall connect timeout
+    defaultConnectTimeout = 30s
+
+    # The default outcall read timeout
+    defaultReadTimeout = 5m
+
     # Set this value to 0 to disable blacklisting.
     connectTimeoutBlacklistDuration = 10s
+
+    # Provides system wide settings for time-critical requests.
```
andyHa, line 130: This is a very coarse and loose description. I'd rather define a proper timeout on the actual outcall itself once created instead of defining yet another global default....

andyHa (changes requested):

andyHa (changes requested):

src/main/java/sirius/kernel/xml/Outcall.java
marvinwoop: The idea was to keep the existing defaults while being able to configure them accordingly in the dependant products systemwide. Therefore you'll not have to call `setConnectTimeout` or `setReadTimeout` before every request.

src/main/java/sirius/kernel/xml/Outcall.java
marvinwoop: Renamed and added soap config plus call in SoapClient as discussed

src/main/resources/component-kernel.conf
```diff
@@ -117,9 +117,25 @@ http.outcall {
...
+    # E.g. for requests which a user is waiting on.
+    interactive {
```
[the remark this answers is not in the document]
marvinwoop: I replaced interactive with soap as SoapClient is in sirius-kernel. Other/additional timeouts can be set in dependant projects.

marvinwoop:
Update: I've changed the fixed `interactive` timeout idea to the suggested flexible solution. In `http.outcall.timeouts.*` one can now set different config blocks and within these the connect and read timeout. This provides one with the option to add custom timeouts and parse and set these to the corresponding outcalls with `withConfiguredTimeout`.

src/main/resources/component-kernel.conf
```diff
@@ -109,17 +109,44 @@ cache {
...
+        # connect timeout, is put onto a "blacklist" and no further connections will be attempted for the given
+        # amount of time as defined here.
+        #
+        # This hopefully prevents "cascading" failures e.g. if a page load itself triggers multiple requests
+        # (think of a shop system which tries to fetch prices for multiple items), the connect timeout might
+        # add up pretty quickly. Therefore, we only try once and then stop trying so that we can render the
+        # remaining part of the site (or at least fail gracefully).
+        #
+        # Set this value to 0 to disable blacklisting.
+        connectTimeoutBlacklistDuration = 10s
```
qw3ry, line 130: Why not scope the blacklist duration as well?

qw3ry: ... or put it outside of `timeouts`; it is actually not a timeout, but the blacklisting duration. Which also would avoid breaking existing config

src/main/java/sirius/kernel/xml/Outcall.java
```diff
@@ -325,6 +331,35 @@ public void setReadTimeout(int timeoutMillis) {
...
+     * See the http.outcall.timeouts.soap block in component-kernel.conf for reference.
+     *
+     * @param configKey the config key of the timeout configuration block
+     * @return this for fluent method calls
+     */
+    public Outcall withConfiguredTimeout(@Nonnull String configKey) {
+        Duration connectTimeout =
+                Sirius.getSettings().getDuration(Strings.apply("http.outcall.timeouts.%s.connectTimeout", configKey));
+        Duration readTimeout =
+                Sirius.getSettings().getDuration(Strings.apply("http.outcall.timeouts.%s.readTimeout", configKey));
```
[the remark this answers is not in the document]
marvinwoop: According to my tests it is working as expected. In which case wouldn't it work so that `getExtension` would be preferable?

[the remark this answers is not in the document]
marvinwoop: As discussed. Config is resolved correspondingly and will be put together so that `getDuration` will get the most recently overwritten changes.

src/main/resources/component-kernel.conf
marvinwoop: the black list works on url basis. But config key was moved to make pr non-breaking.

qw3ry (approved):

jmuscireum (approved):

sabieber (approved):

fhaScireum (approved):

idlira (approved):

src/main/java/sirius/kernel/xml/Outcall.java
```diff
@@ -335,11 +332,32 @@ public void setReadTimeout(int timeoutMillis) {
...
+     * See the http.outcall.timeouts.soap block in component-kernel.conf for reference.
+     *
+     * @param configKey the config key of the timeout configuration block
+     * @return this for fluent method calls
      */
-    public void setInteractive() {
-        setConnectTimeout((int) interactiveConnectTimeout.toMillis());
-        setReadTimeout((int) interactiveReadTimeout.toMillis());
+    public Outcall withConfiguredTimeout(@Nonnull String configKey) {
+        Duration connectTimeout =
```
andyHa, line 344: We should probably get an extension like ext = Sirius.getSettings().getExtension("http.outcall.timeouts", configKey);
ext.get(...) 

Thisway, defaults of the default section will apply and the check below can be removed

marvinwoop: suggestions are now implemented

src/main/java/sirius/kernel/xml/Outcall.java
```diff
@@ -341,19 +342,16 @@ public void setReadTimeout(int timeoutMillis) {
...
-                Sirius.getSettings().getDuration(Strings.apply("http.outcall.timeouts.%s.connectTimeout", configKey));
-        Duration readTimeout =
-                Sirius.getSettings().getDuration(Strings.apply("http.outcall.timeouts.%s.readTimeout", configKey));
-
-        if (!connectTimeout.isZero() && !readTimeout.isZero()) {
-            setConnectTimeout((int) connectTimeout.toMillis());
-            setReadTimeout((int) readTimeout.toMillis());
+        Extension extension = Sirius.getSettings().getExtension("http.outcall.timeouts", configKey);
+
+        if (extension != null) {
```
andyHa, line 347: this cannot be null as a default is present

src/main/resources/component-kernel.conf
```diff
@@ -126,6 +126,8 @@ http.outcall {
     # This timeouts can be expanded by new blocks in dependant projects. For the configuration to take effect, on an
     # outcall the sirius.kernel.xml.Outcall.withConfiguredTimeout needs to be called with the defined block
     # key as parameter.
+    #
+    # NOTE: Please make sure to always define connect- and readTimeout in new blocks to avoid provoking a ConfigException.
```
andyHa, line 130: ? defaults should work independent of each other, so an extension may only define a connect  timeout or vice versa

andyHa (changes requested):

src/main/resources/component-kernel.conf
marvinwoop: You're right, i tested the wrong way. It works as it should.

andyHa (approved):

PR #380, opened 2022-02-11
Restores the original follow-redirects behavior of Outcall and improves error handling.

andyHa:
-> We now follow "normal" redirects by default
-> We now throw an IOException if the body of a non-OK response is being accessed.

Fixes: SIRI-533

### DIFF
--- a/src/main/java/sirius/kernel/xml/Outcall.java
+++ b/src/main/java/sirius/kernel/xml/Outcall.java
@@ -369,6 +369,10 @@ public class Outcall {
 
     /**
      * Executes the outcall and returns the response.
+     * <p>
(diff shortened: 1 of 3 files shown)


src/main/java/sirius/kernel/xml/Outcall.java
```diff
@@ -378,55 +317,54 @@ public Outcall withConfiguredTimeout(@Nonnull String configKey) {
     }
 
     /**
-     * Returns the response header with the given name.
+     * Sents the given context as POST to the designated server.
```
sabieber, line 320:
```suggestion
     * Sends the given context as POST to the designated server.
```

sabieber (approved):

PR #383, opened 2022-02-23
Adds own redirect logic to support unencoded location headers

meg1502:
For example, if a location header contains a space in its path, if we let http client handle the redirects, it will abort with an IllegalArgumentException. Therefore, we no longer instruct the client to follow redirects, but redirect ourselves to support these headers.

Fixes: SIRI-479

Not to reviewers: We use the raw ints for response codes instead of constants, as the constants are part of netty and only available in sirius-web. The constants in HttpUrlConnection is missing some of the newer redirect codes. Therefore the ints and comments are used.

### DIFF
--- a/src/main/java/sirius/kernel/xml/Outcall.java
+++ b/src/main/java/sirius/kernel/xml/Outcall.java
@@ -11,6 +11,7 @@ package sirius.kernel.xml;
 import sirius.kernel.Sirius;
 import sirius.kernel.async.Operation;
 import sirius.kernel.commons.Context;
+import sirius.kernel.commons.Explain;
 import sirius.kernel.commons.Monoflop;
 import sirius.kernel.commons.Streams;
 import sirius.kernel.commons.Strings;
@@ -34,10 +35,14 @@ import java.io.InputStreamReader;
 import java.io.OutputStream;
 import java.io.UnsupportedEncodingException;
 import java.net.ConnectException;
+import java.net.MalformedURLException;
 import java.net.SocketTimeoutException;
 import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
 import java.net.URLEncoder;
 import java.net.http.HttpClient;
+import java.net.http.HttpHeaders;
 import java.net.http.HttpRequest;
 import java.net.http.HttpResponse;
 import java.net.http.HttpTimeoutException;
@@ -79,6 +84,7 @@ public class Outcall {
     private static final String CONTENT_TYPE_FORM_URLENCODED = "application/x-www-form-urlencoded; charset=utf-8";
     private static final Pattern CHARSET_PATTERN = Pattern.compile("(?i)\\bcharset=\\s*\"?([^\\s;\"]*)");
     private static final X509TrustManager TRUST_SELF_SIGNED_CERTS = new TrustingSelfSignedTrustManager();
+    private static final int MAX_REDIRECTS = 5;
 
     /**
      * Keeps track of hosts for which we ran into a connect timeout.
@@ -109,6 +115,7 @@ public class Outcall {
     private final HttpClient.Builder clientBuilder;
     private final HttpRequest.Builder requestBuilder;
     private HttpResponse<InputStream> response;
+    private HttpClient.Redirect redirectPolicy = HttpClient.Redirect.NORMAL;
     private Charset charset = StandardCharsets.UTF_8;
 
     // Provide an output stream for old apis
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/kernel/xml/Outcall.java
```diff
@@ -422,6 +427,14 @@ private void connect() throws IOException {
             request = requestBuilder.build();
         }
 
+        int connections = 0;
```
andyHa, line 430: attempts?

src/main/java/sirius/kernel/xml/Outcall.java
```diff
@@ -422,6 +427,14 @@ private void connect() throws IOException {
...
+        do {
+            makeRequest();
+            connections++;
+        } while (connections < MAX_REDIRECTS && checkAndBuildRedirectRequest(response));
+    }
```
andyHa, line 435: I'd initialize attempts = MAX_REDIRECTS and then use a while(attempts-- > 0) {..} loop, as these are simpler to reason about.

also I'd rename makeRequest to performRequest and let it return the status code. This code could then be used to check if another repetition is needed. This would illustrate the controlflow a bit:
```
while(attempts-- > 0) {
   int statusCode = performRequest();
   if (redirectMode != NEVER && redirectReceived(statusCode)) {
      adaptRequestToRedirect();
   } else {
      return;
   }
}
```

src/main/java/sirius/kernel/xml/Outcall.java
```diff
@@ -626,4 +639,112 @@ public HttpRequest getRequest() {
...
+            // 306: Unused => don't follow
+            case 300, 304, 305, 306 -> false;
+            // 301: Moved permanently => follow
+            // 302: Found => follow
+            // 303: See other => follow
+            // 307: Temp Redirect => follow
+            // 308: Permanent Redirect => follow
+            default -> true;
+        };
+    }
```
andyHa, line 694: Couldn't this be way simpler with an if code == 301 || code == 302 || code == 303 || code == 307 || code == 308

Also URLConnection provides constants for all codes..... (Maybe the HttpClient does as well....)

src/main/java/sirius/kernel/xml/Outcall.java
```diff
@@ -626,4 +639,112 @@ public HttpRequest getRequest() {
...
+
+    @SuppressWarnings("DuplicateBranchesInSwitch")
+    @Explain("Duplicated case is there to visualize expected values.")
+    private boolean shouldSwitchToGet(int statusCode, String originalMethod) {
+        return switch (statusCode) {
+            case 301, 302 -> "POST".equals(originalMethod);
+            case 303 -> true;
+            case 307, 308 -> false;
+
+            default -> false; // unexpected
```
andyHa, line 704:
```suggestion
            default -> false;
```

src/main/java/sirius/kernel/xml/Outcall.java
```diff
@@ -626,4 +639,112 @@ public HttpRequest getRequest() {
...
+        };
+    }
+
+    private URI makeRedirectedURI(HttpHeaders headers) throws IOException {
+        Optional<String> locationHeader = headers.firstValue("Location");
```
andyHa, line 709: Isn't a constant around for this?

src/main/java/sirius/kernel/xml/Outcall.java
```diff
@@ -626,4 +639,112 @@ public HttpRequest getRequest() {
...
+        if (locationHeader.isEmpty()) {
+            throw new ConnectException("Invalid redirection");
+        }
+        URI redirectedURI = makeURIFromLocation(locationHeader.get());
+
+        // redirect could be relative to original URL
+        redirectedURI = request.uri().resolve(redirectedURI);
+        return redirectedURI;
```
andyHa, line 717:
```suggestion
        return request.uri().resolve(redirectedURI);
```

src/main/java/sirius/kernel/xml/Outcall.java
```diff
@@ -626,4 +639,112 @@ public HttpRequest getRequest() {
...
+            default -> false; // unexpected
+        };
+    }
+
+    private URI makeRedirectedURI(HttpHeaders headers) throws IOException {
+        Optional<String> locationHeader = headers.firstValue("Location");
+        if (locationHeader.isEmpty()) {
+            throw new ConnectException("Invalid redirection");
+        }
+        URI redirectedURI = makeURIFromLocation(locationHeader.get());
```
andyHa, line 713:
```suggestion
        URI redirectedURI = headers.firstValue("Location").orElseThrow(() ->   throw new ConnectException("Invalid redirection")).map(this::makeURIFromLocation));
```

andyHa (changes requested):

src/main/java/sirius/kernel/xml/Outcall.java
idlira: or put these codes in a list and check with contains?

src/main/java/sirius/kernel/xml/Outcall.java
meg1502: .map can't be easily done with a function that has a throws declaration

andyHa (approved):

idlira (approved):

mkeckmkeck (approved):

PR #384, opened 2022-03-03
Prevents NPE when null is passed to Strings.urlDecode

cschierle:


idlira (commented):
Not sure if the caller should track this...
Anyway, then the method's return should be marked as nullable too

sabieber (approved):

sabieber:
I think giving the nullable value back to the caller is in line with the other `Strings` method but yes, marking the return value as `Nullable` would be nice

cschierle:
> Not sure if the caller should handle this... Anyway, then the method's return should be marked as nullable too

`urlDecode` now behaves the same as `urlEncode`, which already had a proper null-check.

Added the `Nullable` annotations.

sabieber (approved):

mkeckmkeck (approved):

idlira (approved):

PR #385, opened 2022-03-11
Makes getting date formatters more robust

cschierle:
Fixes: SIRI-525

### DIFF
--- a/src/main/java/sirius/kernel/commons/Value.java
+++ b/src/main/java/sirius/kernel/commons/Value.java
@@ -1329,6 +1329,9 @@ public class Value {
      * @return the parsed LocalDate wrapped as a Value, or the Value itself if it was empty or a parsing error occurred
      */
     public Value tryParseLocalDate(String languageCode) {
+        if (Strings.isEmpty(languageCode)) {
+            return this;
+        }
         return tryParseLocalDate(NLS.getDateFormat(languageCode));
     }
 
--- a/src/main/java/sirius/kernel/commons/Value.java
+++ b/src/main/java/sirius/kernel/commons/Value.java
@@ -1304,10 +1304,11 @@ public class Value {
      * Parses the value as {@link LocalDate} using the given {@link DateTimeFormatter formatter}.
      *
      * @param formatter the formatter for the expected date format
-     * @return the parsed LocalDate wrapped as a Value, or the Value itself if it was empty or a parsing error occurred
+     * @return the parsed LocalDate wrapped as a Value on success. Returns the Value itself, if it or the given
+     * formatter were empty, or if a parsing error occurred.
      */
     public Value tryParseLocalDate(DateTimeFormatter formatter) {
-        if (isNull()) {
+        if (isNull() || formatter == null) {
             return this;
         }
         LocalDate localDate = asLocalDate(null);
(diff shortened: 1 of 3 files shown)


sabieber (approved):

src/main/java/sirius/kernel/nls/NLS.java
```diff
@@ -667,10 +663,16 @@ public static DateTimeFormatter getDateTimeFormat(String lang) {
...
-                                                          "NLS.patternDateTime.withoutSeconds",
-                                                          l)));
+        return safeGetDateTimeFormat(lang, dateTimeFormatters, "NLS.patternDateTime.withoutSeconds");
+    }
+
+    private static DateTimeFormatter safeGetDateTimeFormat(String lang,
+                                                           Map<String, DateTimeFormatter> formatters,
+                                                           String nlsProperty) {
+        if (Strings.isEmpty(lang)) {
+            lang = getCurrentLang();
```
andyHa, line 673: assigning to parameters is commonly not the preferred choice - we could just return here...

andyHa, line 673:
```suggestion
             return formatters.computeIfAbsent(getCurrentLang(), language -> DateTimeFormatter.ofPattern(get(nlsProperty, language)));
```

src/main/java/sirius/kernel/nls/NLS.java
```diff
@@ -667,10 +663,16 @@ public static DateTimeFormatter getDateTimeFormat(String lang) {
...
+        }
+        return formatters.computeIfAbsent(lang, l -> DateTimeFormatter.ofPattern(get(nlsProperty, l)));
```
andyHa, line 675:
```suggestion
        return formatters.computeIfAbsent(lang, ignored -> DateTimeFormatter.ofPattern(get(nlsProperty, lang)));
```

src/main/java/sirius/kernel/commons/Value.java
```diff
@@ -1329,6 +1329,9 @@ public Value tryParseLocalDate(DateTimeFormatter formatter) {
      * @return the parsed LocalDate wrapped as a Value, or the Value itself if it was empty or a parsing error occurred
      */
     public Value tryParseLocalDate(String languageCode) {
+        if (Strings.isEmpty(languageCode)) {
```
andyHa, line 1332: in every other location, we'd use getCurrentLang() here ?! is this consistent?

andyHa (changes requested):

src/main/java/sirius/kernel/commons/Value.java
cschierle: This method only **tries** to parse the value as a LocalDate. So when an empty language is passed, there is no need to try any further.

I think for a `tryParseLocalDate` returning without defaulting to the current language is the expected behavior. If this was a `parseLocalDate`, then the default should be used.

fhaScireum (approved):

idlira (approved):

andyHa (approved):

src/main/java/sirius/kernel/commons/Value.java
andyHa: Still in all other methods "null" is used to represent the default language. And the try is meant to "tryParse" not to check parameters(?). I'd say if parseLocalDate yields a value, tryParseLocalDate with **the same parameters** should also return a value.

andyHa (approved):

PR #386, opened 2022-04-05
Makes default headers and timeouts accessible

cschierle:
Fixes: OX-8364


andyHa (approved):

src/main/java/sirius/kernel/xml/Outcall.java
```diff
@@ -76,10 +76,10 @@
 
     private static final String REQUEST_METHOD_HEAD = "HEAD";
     private static final String HEADER_CONTENT_TYPE = "Content-Type";
-    private static final String HEADER_USER_AGENT = "User-Agent";
-    private static final String HEADER_ACCEPT = "Accept";
+    public static final String HEADER_USER_AGENT = "User-Agent";
```
idlira, line 79: Maybe just rearrange them so public vs privates won't stay mixed?

idlira (approved):

PR #390, opened 2022-05-02
Provides a more elaborate way of translating config key/values.

andyHa:


src/test/java/sirius/kernel/settings/SettingsTest.kt
```diff
@@ -0,0 +1,44 @@
...
+            directKey = "test"
+            intKey = 5
+            translatedKey = "${'$'}testKey"
+            mapKey {
+                en: test
+            }
+            
+            mapKeyWithDefault {
+                default: test
+            }
```
sabieber, line 31: Maybe combine into case with lang+default and an assert for the language and a missing language (to trigger default/fallback). This would match the real world a bit closer.

src/main/java/sirius/kernel/settings/Settings.java
```diff
@@ -267,18 +269,77 @@ public Duration getDuration(String path) {
...
+        Value value = get(key);
+        if (value.isFilled() && value.is(String.class)) {
+            return NLS.smartGet(value.asString(), lang);
+        }
+
+        if (value.is(Map.class)) {
+            Map<String, String> translations = value.get(Map.class, Collections.emptyMap());
+            String effectiveLanguage = Strings.isEmpty(lang) ? NLS.getCurrentLang() : lang;
+            return Optional.ofNullable(translations.get(effectiveLanguage))
+                           .or(() -> Optional.ofNullable(translations.get("default")))
```
sabieber, line 318: Should we name this "fallback" so we don't have to do any mappings/transformations when storing a MultiLanguageString to a config (via UI)?

sabieber (approved):

src/main/java/sirius/kernel/settings/Settings.java
qw3ry: Do we have any `Config` -> `String` transformation anywhere?
Naming it "default" makes perfekt sense to me, as the fallbacks in `Extension`s are also called "default"

qw3ry (approved):

src/main/java/sirius/kernel/commons/Value.java
```diff
@@ -1943,9 +1946,12 @@ public Value translate() {
      * if the string representation of the wrapped value starts with {@code $}.
      * The dollar sign is skipped when passing the key to <tt>NLS</tt>. Otherwise <tt>this</tt> is returned.
      * @see NLS#get(String, String)
+     * @deprecated User {@link sirius.kernel.settings.Settings#getTranslatedString(String, String)} if this value is
+     * loaded from a Config/Settings. Otherwise use {@link NLS#smartGet(String, String)}
      */
     @Nonnull
     @CheckReturnValue
+    @Deprecated
```
idlira, line 1954: Replace usages within kernel otherwise we can't build without accepting deprecations.
`[WARNING] /drone/src/src/main/java/sirius/kernel/settings/Extension.java:[80,27] [deprecation] translate() in Value has been deprecated.`

idlira (changes requested):

meg1502 (approved):

mkeckmkeck (approved):

src/main/java/sirius/kernel/settings/Settings.java
andyHa: For extensions we did us "default" so it felt more natural than using fallback. However, as these things are unrelated, we can also use fallback.

idlira (approved):

src/main/java/sirius/kernel/commons/Value.java
```diff
@@ -1927,9 +1927,12 @@ public int hashCode() {
      * if the string representation of the wrapped value starts with {@code $}.
      * The dollar sign is skipped when passing the key to <tt>NLS</tt>. Otherwise <tt>this</tt> is returned.
      * @see NLS#get(String)
+     * @deprecated User {@link sirius.kernel.settings.Settings#getTranslatedString(String)} if this value is
```
cschierle, line 1930:
```suggestion
     * @deprecated Use {@link sirius.kernel.settings.Settings#getTranslatedString(String)} if this value is
```

src/main/java/sirius/kernel/commons/Value.java
```diff
@@ -1943,9 +1946,12 @@ public Value translate() {
      * if the string representation of the wrapped value starts with {@code $}.
      * The dollar sign is skipped when passing the key to <tt>NLS</tt>. Otherwise <tt>this</tt> is returned.
      * @see NLS#get(String, String)
+     * @deprecated User {@link sirius.kernel.settings.Settings#getTranslatedString(String, String)} if this value is
```
cschierle, line 1949:
```suggestion
     * @deprecated Use {@link sirius.kernel.settings.Settings#getTranslatedString(String, String)} if this value is
```

ymo-sci (approved):

meg1502 (approved):

sabieber (approved):

PR #401, opened 2022-07-25
NLS Bugfix

jakobvogel:
Fixes method to also consider the fallback language before retiring with the key as value.

### DIFF
--- a/src/main/java/sirius/kernel/nls/NLS.java
+++ b/src/main/java/sirius/kernel/nls/NLS.java
@@ -340,9 +340,9 @@ public class NLS {
             return Optional.empty();
         }
         String targetLanguage = Strings.isEmpty(lang) ? getCurrentLang() : lang;
-        if (translation.hasTranslation(targetLanguage)) {
-            return Optional.of(translation.translate(Strings.isEmpty(lang) ? getCurrentLang() : lang,
-                                                     getFallbackLanguage()));
+        String fallbackLanguage = getFallbackLanguage();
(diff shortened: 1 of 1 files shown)


idlira (approved):

src/main/java/sirius/kernel/nls/NLS.java
```diff
@@ -340,9 +340,9 @@ public static Optional<String> getIfExists(String property, @Nullable String lan
             return Optional.empty();
         }
         String targetLanguage = Strings.isEmpty(lang) ? getCurrentLang() : lang;
-        if (translation.hasTranslation(targetLanguage)) {
-            return Optional.of(translation.translate(Strings.isEmpty(lang) ? getCurrentLang() : lang,
-                                                     getFallbackLanguage()));
+        String fallbackLanguage = getFallbackLanguage();
```
andyHa, line 343: maybe only fetch this if needed (split into two checks) - as otherwise we perform a threadlocal lookup for no reason (which is super fast - but super unneccessarry in this case :-) )

sabieber (approved):

qw3ry (approved):

cschierle (approved):
Should the fallbackLanguage be mentioned in the JavaDoc, too?

src/main/java/sirius/kernel/nls/NLS.java
jakobvogel: done

andyHa (approved):

PR #402, opened 2022-09-01
Permits to lookup a named part by its actual class.

andyHa:


src/main/java/sirius/kernel/di/GlobalContext.java
```diff
@@ -53,6 +54,20 @@ public interface GlobalContext {
     @Nullable
     <P> P getPart(@Nonnull String uniqueName, @Nonnull Class<P> clazz);
 
+    /**
+     * Returns the instance of <tt>implementationClass</tt> which has been registered for <tt>lookupClass</tt>.
+     * <p>
+     * This is more of an edge case, were we know that an instance of a given class has been registered, but not for
+     * its own class, but (at least) for the given <tt>lookupClass</tt>.
```
jakobvogel, line 61:
```suggestion
     * This is more of an edge case, where we know that an instance of a given class has been registered, but not for
     * its own class, but (at least) for the given <tt>lookupClass</tt>.
```

src/main/java/sirius/kernel/di/GlobalContext.java
```diff
@@ -53,6 +54,20 @@ public interface GlobalContext {
...
+     *
+     * @param lookupClass         the class for which the requested part has been registered
+     * @param implementationClass the actual class of the part to fetch
+     * @param <P>                 the generic type of the implementation class
+     * @return the registered instance ir <tt>null</tt> if none was present
```
jakobvogel, line 66:
```suggestion
     * @return the registered instance or <tt>null</tt> if none was present
```

jakobvogel (approved):

sabieber (approved):

PR #408, opened 2022-10-06
Uses the correct date time format for the If-Modified-Since header

jmuscireum:
Apache expects the day to use two digits as specified by RFC 7231 (section 7.1.1.1.) and will ignore the header if the incorrect date format is used.

### DIFF
--- a/src/main/java/sirius/kernel/xml/Outcall.java
+++ b/src/main/java/sirius/kernel/xml/Outcall.java
@@ -57,6 +57,7 @@ import java.time.ZoneId;
 import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
 import java.util.Base64;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/kernel/xml/Outcall.java
```diff
@@ -78,6 +81,19 @@ public class Outcall {
     public static final String HEADER_ACCEPT = "Accept";
     public static final String HEADER_ACCEPT_DEFAULT_VALUE = "*/*";
 
+    /**
+     * Date time formatter as per
+     * <a href="https://datatracker.ietf.org/doc/html/rfc9110#section-5.6.7">RFC 9110 section 5.6.7.</a>
+     * <p>
+     * In contrast to {@link DateTimeFormatter#RFC_1123_DATE_TIME}, the day must use two digits.
```
andyHa, line 88: also, we not always out put UTC (indicated by the legacy zone id GMT) - 1123 uses the local TZ offset like +0200

sabieber (approved):

mkeckmkeck (approved):

andyHa (approved):

idlira (approved):

sabieber (approved):

ymo-sci (approved):

PR #410, opened 2022-10-07
Switches project infrastructure from Maven to Gradle

cschierle:
Fixes: SIRI-632

### Breaking changes:
- Maven setup (e.g. `pom.xml`) has been replaced by Gradle equivalent
- Drone config in `drone.yml` has been updated to use Gradle tasks for validation and publication


build.gradle.kts
```diff
@@ -31,6 +31,8 @@ dependencies {
 
     // Provides the tools to load the system configuration
     api("com.typesafe:config:1.4.1")
+    // scireum parsii
+    api("com.scireum:parsii:5.0.1")
```
sabieber, line 33: Glaub die kann jetzt wieder raus wenn der aktuelle develop reingemerged wurde

sabieber (approved):

build.gradle.kts
cschierle: stimmt, guter Punkt. Das ist ja mittlerweile auch in kernel gewandert

sabieber (approved):

mkeckmkeck (approved):
Will plain junit test written in Java still be executed?

cschierle:
> Will plain junit test written in Java still be executed?

Yes! The intended setup is now as follows:
- Any old tests written in Groovy (usually named `...Spec.groovy`) are executed via TestSuite and use the JUnit 4 Test Runner
- Any JUnit 5 tests written in Java or Kotlin (named `...Test.java/.kt`) are executed directly with the JUnit Jupiter framework and start/stop the sirius environment, if the test class is annotated with `@ExtendWith(SiriusExtension)`.

For example [LimitTest.java](https://github.com/scireum/sirius-kernel/blob/feature/csc/GradleWithNewTests/src/test/java/sirius/kernel/commons/LimitTest.java) is still executed

src/test/java/sirius/kernel/SiriusExtension.java
```diff
@@ -30,4 +32,9 @@ public void beforeAll(ExtensionContext context) {
     public void beforeEach(ExtensionContext context) throws Exception {
         CallContext.initialize();
     }
+
+    @Override
+    public void afterAll(ExtensionContext context) throws Exception {
+        TestHelper.tearDown(SiriusExtension.class);
```
mkeckmkeck, line 38: Framework starts 3 times. I suppose shutting it down here is wrong. What do we need that for?
https://drone.scireum.com/scireum/sirius-kernel/2905/1/2

mkeckmkeck (changes requested):

src/test/java/sirius/kernel/SiriusExtension.java
mkeckmkeck: Also, is there no surefire collecting test statistics?

idlira (approved):

src/test/java/sirius/kernel/SiriusExtension.java
cschierle: Thanks for this improvement.

We still have to start the framework once for the Groovy Specs run via Scenario Runner, and a second time for all JUnit 5 Tests though. Instead of optimizing this further, we should try and get rid of the Groovy specs completely.

Test statistics are now also printed

PR #412, opened 2022-10-12
Adds new methods for safely adding path parts to an URL builder

sabieber:
Fixes: [SIRI-641](https://scireum.myjetbrains.com/youtrack/issue/SIRI-641)

### DIFF
--- a/src/main/java/sirius/kernel/commons/URLBuilder.java
+++ b/src/main/java/sirius/kernel/commons/URLBuilder.java
@@ -8,12 +8,15 @@
 
 package sirius.kernel.commons;
 
+import org.apache.commons.lang3.StringUtils;
+
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
+import java.util.Arrays;
 
 /**
  * Used to successively build URLs.
@@ -33,12 +36,13 @@ public class URLBuilder {
     public static final String PROTOCOL_HTTPS = "https";
 
     private static final String QUERY_SEPARATOR = "?";
+    private static final String PATH_SEPARATOR = "/";
 
     private final StringBuilder url;
     private final Monoflop questionMark = Monoflop.create();
 
     /**
-     * Creates a new instance pre filled with the given baseURL.
+     * Creates a new instance pre-filled with the given baseURL.
      *
      * @param baseURL the base url to stat with in a form like <tt>http://somehost.com</tt> or
      *                <tt>http://somehost.com?some=parameter</tt>
@@ -50,7 +54,7 @@ public class URLBuilder {
             questionMark.toggle();
         }
 
-        if (baseURL.endsWith("/")) {
+        if (baseURL.endsWith(PATH_SEPARATOR)) {
             url.append(baseURL, 0, baseURL.length() - 1);
         } else {
             url.append(baseURL);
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/kernel/commons/URLBuilder.java
```diff
@@ -70,6 +74,55 @@ public URLBuilder(@Nonnull String protocol, @Nonnull String host) {
         url.append(host);
     }
 
+    /**
+     * Adds a path part to the url.
```
jakobvogel, line 78: Sollten wir das als Abkürzung nicht groß schreiben?

```suggestion
     * Adds a path part to the URL.
```

src/main/java/sirius/kernel/commons/URLBuilder.java
```diff
@@ -70,6 +74,55 @@ public URLBuilder(@Nonnull String protocol, @Nonnull String host) {
...
+     * <p>
+     * Once the first parameter has been added, the path can no longer be modified. Also, the part itself can (but
+     * shouldn't) contain parameters, usually led by an initial question mark.
+     *
+     * @param uriPartToAdd the uri part to add. Included trailing or leading '/' are handled automatically.
```
jakobvogel, line 83: Dto.

```suggestion
     * @param uriPartToAdd the URI part to add. Included trailing or leading '/' are handled automatically.
```

src/main/java/sirius/kernel/commons/URLBuilder.java
```diff
@@ -70,6 +74,55 @@ public URLBuilder(@Nonnull String protocol, @Nonnull String host) {
...
+     *                     If this contains a '?' to add parameters, no more parts can be added afterwards.
+     * @return the builder itself for fluent method calls
+     */
+    public URLBuilder addSafePart(@Nullable String uriPartToAdd) {
+        if (Strings.isEmpty(uriPartToAdd)) {
+            return this;
+        }
+
+        if (questionMark.isToggled()) {
+            throw new IllegalStateException(Strings.apply("Cannot add '%s'! Parameters where already added to: '%s'.",
```
idlira, line 93:
```suggestion
            throw new IllegalStateException(Strings.apply("Cannot add '%s'! Parameters were already added to: '%s'.",
```

idlira (approved):

jakobvogel (approved):

sabieber:
Thanks for your suggestions @idlira & @jakobvogel but had to apply them to the whole class as it was wrong everywhere and I just oriented myself at the other existing methods 😉

idlira (approved):

andyHa (approved):

qw3ry:
If we want to remove `addPart`, why not just make the PR breaking and change its semantics? Once `addPart` is gone, the remaining `addSafePart` will sound somewhat awkward, and clumsy (without an unsafe counterpart)

sabieber:
> If we want to remove `addPart`, why not just make the PR breaking and change its semantics? Once `addPart` is gone, the remaining `addSafePart` will sound somewhat awkward, and clumsy (without an unsafe counterpart)

The idea was that it is much easier and safer to detect where the old method is called with semantics that are no longer supported (as seen in the tests or here: https://github.com/scireum/oxomi/pull/9061/files#diff-7be4d3f73bf8a93ed7740720be6f70c780d76dbdaa5e11713dbc619b096ab1a8R255)

Also it was discussed to maybe at a later date add a `addRawPart` if needed, that skips all the safety checks.

qw3ry (approved):

mkeckmkeck (approved):

PR #416, opened 2022-11-22
Improves output of everyday timers in timer command

sabieber:
By sorting the list (according to the hour of day the tasks are executed at) and switching the hour and name around.

Before:

image

After:

image


meg1502 (approved):

jakobvogel (approved):
und auch noch ne for-schleife losgeworden 🤩

PR #417, opened 2022-11-22
More stable wiring / system startup.

andyHa:
If a class cannot be loaded as a referenced class is missing, we simply ignore the class instead of crashing the startup.

This might be a class for a disabled framework which misses its driver or the like.

We now gracefully output errors during static wiring (which are pretty much expected if drivers are omitted):
grafik

but we report a proper warning, if the questionable part is enabled and actively used:
grafik

### DIFF
--- a/src/main/java/sirius/kernel/di/PartRegistry.java
+++ b/src/main/java/sirius/kernel/di/PartRegistry.java
@@ -16,7 +16,6 @@ import sirius.kernel.commons.Strings;
 import sirius.kernel.commons.Tuple;
 import sirius.kernel.di.std.Named;
 import sirius.kernel.di.std.Priorized;
-import sirius.kernel.health.Exceptions;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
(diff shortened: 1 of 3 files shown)


jakobvogel (approved):

src/main/java/sirius/kernel/di/PartRegistry.java
```diff
@@ -148,30 +149,39 @@ void wireClass(Class<?> clazz) {
...
-                                                                        + "(%s)",
-                                                                        p.getTrigger().getName(),
-                                                                        field.getDeclaringClass().getName(),
-                                                                        field.getName(),
-                                                                        e.getMessage(),
-                                                                        e.getClass().getName());
-                                                            }
-                                                        });
+        try {
+            for (Field field : clazz.getDeclaredFields()) {
```
mkeckmkeck, line 153: Doesn't this skip all following wirings when an error occurs?

mkeckmkeck (changes requested):

src/main/java/sirius/kernel/di/PartRegistry.java
```diff
@@ -148,30 +149,39 @@ void wireClass(Class<?> clazz) {
...
+                if (!Modifier.isFinal(field.getModifiers()) && (object != null
+                                                                || Modifier.isStatic(field.getModifiers()))) {
+                    wireField(object, field);
+                }
             }
+        } catch (NoClassDefFoundError e) {
+            // These errors probably come from a missing driver or the like and can most probably be ignored...
+            Exceptions.ignore(e);
```
qw3ry, line 161: Better log the exception. IMO this should only be a "quick fix" so the application can actually start at all. We should really handle these cases with proper framework handling.

sabieber: Then every startup of systems without the IBM driver will log this at every startup?

sabieber (approved):

src/main/java/sirius/kernel/di/PartRegistry.java
mkeckmkeck: Nevermind, as one field won't be filled it is ok to break here.

mkeckmkeck (approved):
Would prefer a customization approach, but works for me

qw3ry (approved):

PR #423, opened 2023-01-02
Deprecates bad methods for outcall and enhance existing method getInput 

ymo-sci:
The idea was to make getInput return the XML response (if there was one) instead of failing for a non 200 status. We therefore check the content-type and only fail, if a non XML response is present. This is more in-line with the responses generated by @PublicService in sirius-web.

### DIFF
--- a/src/main/java/sirius/kernel/xml/XMLCall.java
+++ b/src/main/java/sirius/kernel/xml/XMLCall.java
@@ -135,7 +135,8 @@ public class XMLCall {
      * @throws IOException in case of an IO error while receiving the result
      */
     public XMLStructuredInput getInput() throws IOException {
-        if (!outcall.isErroneous() || outcall.getHeaderField("content-type").toLowerCase().contains("xml")) {
+        String contentType = outcall.getHeaderField("content-type");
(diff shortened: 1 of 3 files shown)


andyHa (approved):

meg1502 (approved):

sabieber (approved):

mkeckmkeck (approved):

fhaScireum (approved):

src/main/java/sirius/kernel/xml/XMLCall.java
```diff
@@ -134,16 +135,10 @@ public XMLStructuredOutput getOutput() throws IOException {
...
-
-    /**
-     * Provides access to the raw answer of the call.
-     *
-     * @return the result of the call
-     * @throws IOException in case of an IO error while receiving the result
-     */
-    public XMLStructuredInput getRawInput() throws IOException {
-        return new XMLStructuredInput(outcall.getResponse().body(), namespaceContext);
+        if (!outcall.isErroneous() || outcall.getHeaderField("content-type").toLowerCase().contains("xml")) {
```
cschierle, line 138: `getHeaderField` is nullable! Maybe check if the header is present first, then check if it contains `xml`

cschierle (approved):

PR #433, opened 2023-02-08
Adds javadoc to method to document 'magic' behavior

mko-sci:
Fixes: [OX-9538](https://scireum.myjetbrains.com/youtrack/issue/OX-9538)


andyHa:
Maybe we should also document tne magic parameters errorMessage and errorClass for the i18n version (withFormattedMessage - or the like), while we're at it?

src/main/java/sirius/kernel/health/Exceptions.java
```diff
@@ -114,6 +114,8 @@ protected ErrorHandler(boolean processError) {
 
         /**
          * Specifies which exception leaded to the error being handled.
+         * <p>
+         * Note: When used, an additional '- %s (%s)' has to be added at the end of the error message provided when concatenating a {@link ErrorHandler#withSystemErrorMessage(String, Object...)} call.
```
mkeckmkeck, line 118: Better add this to the withSystemErrorMessage call? Also, outputting the additional parameters is optional.

mko-sci:
Gibts noch mehr stellen wo das Comment net schaden würde?

sabieber (approved):

mkeckmkeck (approved):

src/main/java/sirius/kernel/health/Exceptions.java
```diff
@@ -114,6 +114,8 @@ protected ErrorHandler(boolean processError) {
 
         /**
          * Specifies which exception leaded to the error being handled.
+         * <p>
+         * Note: When used, an additional '- %s (%s)' could be added at the end of the error message provided when concatenating a {@link ErrorHandler#withSystemErrorMessage(String, Object...)} call.
```
andyHa, line 118:
```suggestion
         * <p>
         * Note: When {@link ErrorHandler#withSystemErrorMessage(String, Object...)} is used, two extra
         * parameters are available for the string template. The first is the exception message and the second
         * is the exception class. These will be appended after all the user-given parameters. Therefore,
         * one can add <tt>%s (%s)</tt> to the message provided here to output the actual error and
         * exception type if needed.
```

src/main/java/sirius/kernel/health/Exceptions.java
```diff
@@ -167,6 +169,8 @@ public ErrorHandler withDirectMessage(String message) {
          * <p>
          * Still a translated message will be created, which notifies the user about the system error and provides
          * the untranslated error message, generated by this method. These messages should be in english.
+         * <p>
+         * Note: When combining with {@link ErrorHandler#error(Throwable)} an optional '- %s (%s)' could be added at the end of the provided message to output the errorMessage and errorClass.
```
andyHa, line 173:
```suggestion
         * <p>
         * Note: When combining with {@link ErrorHandler#error(Throwable)}, two extra
         * parameters are available for the string template. The first is the exception message and the second
         * is the exception class. These will be appended after all the user-given parameters. Therefore,
         * one can add <tt>%s (%s)</tt> to the message provided here to output the actual error and
         * exception type if needed.
```

sabieber (approved):

qw3ry (approved):

cschierle (approved):

PR #437, opened 2023-02-14
Amount as Number (SIRI-719)

jakobvogel:
> https://scireum.myjetbrains.com/youtrack/issue/SIRI-719

`Amount` becomes a `Number`, yay!

### DIFF
--- a/src/main/java/sirius/kernel/commons/Amount.java
+++ b/src/main/java/sirius/kernel/commons/Amount.java
@@ -246,37 +246,25 @@ public class Amount extends Number implements Comparable<Amount>, Serializable {
 
     @Override
     public int intValue() {
-        if (value == null) {
-            throw new IllegalStateException("The value is empty.");
-        }
-
+        throwExceptionIfEmpty();
         return value.intValue();
     }
 
     @Override
     public long longValue() {
-        if (value == null) {
-            throw new IllegalStateException("The value is empty.");
-        }
-
+        throwExceptionIfEmpty();
         return value.longValue();
     }
 
     @Override
     public float floatValue() {
-        if (value == null) {
-            throw new IllegalStateException("The value is empty.");
-        }
-
+        throwExceptionIfEmpty();
         return value.floatValue();
     }
 
     @Override
     public double doubleValue() {
-        if (value == null) {
-            throw new IllegalStateException("The value is empty.");
-        }
-
+        throwExceptionIfEmpty();
         return value.doubleValue();
     }
 
@@ -948,5 +936,11 @@ public class Amount extends Number implements Comparable<Amount>, Serializable {
             return 0;
         }
         return Math.round(Math.floor(Math.log10(value.doubleValue()) + 1));
+    }
+
+    private void throwExceptionIfEmpty() {
+        if (value == null) {
+            throw new IllegalStateException("The value is empty.");
+        }
     }
 }
--- a/src/main/java/sirius/kernel/commons/Amount.java
+++ b/src/main/java/sirius/kernel/commons/Amount.java
@@ -46,7 +46,7 @@ import java.util.function.Supplier;
  * @see BigDecimal
  */
 @Immutable
-public class Amount implements Comparable<Amount>, Serializable {
+public class Amount extends Number implements Comparable<Amount>, Serializable {
 
     @Serial
     private static final long serialVersionUID = 2187873067365153302L;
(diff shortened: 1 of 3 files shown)


src/main/java/sirius/kernel/commons/Value.java
```diff
@@ -893,11 +887,8 @@ public Integer getInteger() {
             if (data instanceof Integer integer) {
```
jakobvogel, line 887: Left this as this is exactly the return type. Ok?

src/main/java/sirius/kernel/commons/Value.java
```diff
@@ -972,8 +954,8 @@ public Long getLong() {
             if (data instanceof Long longValue) {
```
jakobvogel, line 954: Left this as this is exactly the return type. Ok?

src/test/java/sirius/kernel/commons/ValueSpec.groovy
```diff
@@ -168,6 +168,77 @@ class ValueSpec extends BaseSpecification {
...
+        99999.0           | new Long(99999l)
+        new Double(99999) | new Long(99999l)
+        Amount.of(99999)  | new Long(99999l)
+        "99999"           | new Long(99999l)
+        "Keine Zahl"      | null
+    }
+
+    def "Casting to doubles works"() {
+        expect:
+        Value.of(input).asDouble(1234.56) == output
```
jakobvogel, line 229: Is this a valid double comparison in Groovy? There should be an ε

andyHa: Doubles.EPSILON?

andyHa (approved):

src/main/java/sirius/kernel/commons/Amount.java
```diff
@@ -244,6 +244,26 @@ public BigDecimal getAmount() {
         return value;
     }
 
+    @Override
+    public int intValue() {
+        return value.intValue();
```
mkeckmkeck, line 249: Dont't we require null checks + fallback here e.g. for Amount.NOTHING?

idlira (approved):

src/main/java/sirius/kernel/commons/Amount.java
jakobvogel: Hmm, to me it feels ok to have an NPE here when treating an `Amount.NOTHING` as `Number`, but I am open for discussion. For now, I have added such cases to the test, and I have reinforced the casting in `Value`

src/main/java/sirius/kernel/commons/Value.java
sabieber: For simplicity sake I would include it in the Number case and remove this if. When looking at the Implementation of the Integer or Long class they do exactly the same

src/main/java/sirius/kernel/commons/Value.java
sabieber: For simplicity sake I would include it in the Number case and remove this if. When looking at the Implementation of the Integer or Long class they do exactly the same

sabieber (approved):

src/main/java/sirius/kernel/commons/Amount.java
andyHa: Return Integer.MIN_VALUE - this should probably the most unexpected behaviour :-P

andyHa: One could return an IllegalStateException (?) as this would feel less "random" ;-)

sabieber: Right, that's actually a very valid point.

I quite like the idea of an exception, additionally we maybe could expand the filled check methods (if they currently don't have a code path for Amounts) to act like the Value is empty when the Amount is empty

src/main/java/sirius/kernel/commons/Amount.java
```diff
@@ -246,21 +246,37 @@ public BigDecimal getAmount() {
...
+            throw new IllegalStateException("The value is empty.");
+        }
+
         return value.intValue();
     }
 
     @Override
     public long longValue() {
+        if (value == null) {
+            throw new IllegalStateException("The value is empty.");
```
idlira, line 259: Extract in createIllegalStateException? (or define a constant for the string)?

PR #438, opened 2023-02-15
SIRI-724 improve logging

qw3ry:


sabieber (approved):

src/main/java/sirius/kernel/xml/XMLCall.java
```diff
@@ -82,6 +88,19 @@ public static XMLCall to(URI uri, String contentType) throws IOException {
         return result;
     }
 
+    /**
+     * Log the outcall to {@code logger}.
```
qw3ry, line 92:
```suggestion
     * Logs the outcall to {@code logger}.
```

src/main/java/sirius/kernel/async/Tasks.java
```diff
@@ -417,17 +415,16 @@ private void blockUntilExecutorTerminates(String name, AsyncExecutor exec) {
         LOG.INFO("Waiting for async executor '%s' to terminate...", name);
         try {
             if (!exec.awaitTermination(EXECUTOR_SHUTDOWN_WAIT.getSeconds(), TimeUnit.SECONDS)) {
-                LOG.SEVERE(Strings.apply("Executor '%s' did not terminate within 60s. Interrupting " + "tasks...",
-                                         name));
+                LOG.SEVERE("Executor '%s' did not terminate within 60s. Interrupting " + "tasks...", name);
```
jakobvogel, line 418:
```suggestion
                LOG.SEVERE("Executor '%s' did not terminate within 60s. Interrupting tasks...", name);
```

src/main/java/sirius/kernel/health/Log.java
```diff
@@ -352,6 +352,19 @@ public void SEVERE(Object message) {
         log(Level.SEVERE, message);
     }
 
+    /**
+     * Logs the given message at the SEVERE or ERROR level.
+     * <p>
+     * he ERROR level can be used to signal problems or error which occurred in the system. It is recommended to handle
+     * exceptions using {@link Exceptions} instead.
```
jakobvogel, line 359:
```suggestion
    /**
     * Logs the given message at the SEVERE or ERROR level.
     * <p>
     * The ERROR level can be used to signal problems or errors which occurred in the system. It is recommended to handle
     * exceptions using {@link Exceptions} instead.
```

jakobvogel (approved):

src/main/java/sirius/kernel/xml/XMLCall.java
```diff
@@ -82,6 +88,19 @@ public static XMLCall to(URI uri, String contentType) throws IOException {
         return result;
     }
 
+    /**
+     * Logs the outcall to {@code logger}.
+     * <p>
+     * The outcall is only logged when the logger is set to FINE. The default logger is "xml".
+     *
+     * @param logger the logger to log to
+     * @return returns the XML call itself for fluent method calls
```
jakobvogel, line 97:
```suggestion
     * @return the XML call itself for fluent method calls
```

src/main/java/sirius/kernel/xml/XMLCall.java
```diff
@@ -128,16 +147,46 @@ public XMLStructuredOutput getOutput() throws IOException {
...
+                                                                                  
+                                                     ${response}
+                                                     ---------- end ----------
+                                                     """)
+                                     .set("httpMethod", isPost ? "POST" : "GET")
+                                     .set("url", outcall.getRequest().uri())
+                                     .set("callBody", isPost ? getOutput() : null)
+                                     .set("responseCode", getOutcall().getResponseCode())
+                                     .set("response", text)
+                                     .smartFormat());
```
andyHa, line 172: extract method

src/main/java/sirius/kernel/xml/XMLCall.java
```diff
@@ -24,6 +29,7 @@ public class XMLCall {
 
     private Outcall outcall;
     private NamespaceContext namespaceContext;
+    private Log logger = Log.get("xml");
```
andyHa, line 32: fineLogger ? oder debugLog(ger)

src/main/java/sirius/kernel/xml/XMLCall.java
```diff
@@ -82,6 +88,19 @@ public static XMLCall to(URI uri, String contentType) throws IOException {
         return result;
     }
 
+    /**
+     * Logs the outcall to {@code logger}.
```
andyHa, line 92:
```suggestion
     * Logs the request and the resulting response to the given logger using the <tt>FINE</tt> level.
```

src/main/java/sirius/kernel/xml/XMLCall.java
```diff
@@ -128,16 +147,46 @@ public XMLStructuredOutput getOutput() throws IOException {
         return new XMLStructuredOutput(outcall.postFromOutput());
     }
 
+    private InputStream getInputStream() throws IOException {
+        if (logger != null && logger.isFINE()) {
+            // log the request, even when parsing fails
+            try (InputStream body = outcall.getResponse().body()) {
```
andyHa, line 153: is this guaranteed to only execute once? (was we get the body twice...)

andyHa (changes requested):

src/main/java/sirius/kernel/xml/XMLCall.java
```diff
@@ -128,16 +147,46 @@ public XMLStructuredOutput getOutput() throws IOException {
...
+                String text = new String(body.readAllBytes(), StandardCharsets.UTF_8);
```
andyHa, line 154: we need to use the xml parse + transform here as the charset can by anything....

src/main/java/sirius/kernel/xml/XMLCall.java
```diff
@@ -128,16 +147,46 @@ public XMLStructuredOutput getOutput() throws IOException {
...
      *
      * @return the XML result of the call
      * @throws IOException in case of an IO error while receiving the result
      */
     public XMLStructuredInput getInput() throws IOException {
+        InputStream body = getInputStream();
         String contentType = outcall.getHeaderField("content-type");
         if (!outcall.isErroneous() || (contentType != null && contentType.toLowerCase().contains("xml"))) {
-            return new XMLStructuredInput(outcall.getResponse().body(), namespaceContext);
+            return new XMLStructuredInput(body, namespaceContext);
```
andyHa, line 189: remove the extra variable and call in place (?)

andyHa (changes requested):

src/main/java/sirius/kernel/xml/XMLCall.java
qw3ry: Well if you call `#getInput()` multiple times it will probably fail, because the input stream was read before. So it does not matter much if this fails in `#getInput()` or here already, right?

src/main/java/sirius/kernel/xml/XMLCall.java
qw3ry: I would very much like to avoid this - if the response cannot be parsed, nothing can be logged, but these are exactly the cases where a logging would be very helpful

qw3ry: Please have a look at the updated code. I use the encoding suggested by the outcall, and the encoding now only effects the log, not the returned stream

src/main/java/sirius/kernel/xml/XMLCall.java
qw3ry: I'd like to leave it here so failures can be logged as well.

src/main/java/sirius/kernel/xml/XMLCall.java
```diff
@@ -128,16 +146,51 @@ public XMLStructuredOutput getOutput() throws IOException {
...
+                byte[] bytes = body.readAllBytes();
+                logRequest(new String(bytes, outcall.getContentEncoding()));
+                return new ByteArrayInputStream(bytes);
+            }
+        }
+        return outcall.getResponse().body();
+    }
+
+    private void logRequest(String response) throws IOException {
+        boolean isPost = outcall.isPostRequest();
```
mkeckmkeck, line 162: Better use outcall.getRequest().method There is more than one way to post. And more methods than post and get.

mkeckmkeck (approved):

ymo-sci (approved):

idlira (approved):

andyHa (approved):

PR #440, opened 2023-02-21
Enhances parsing of strings into integer/longs

idlira:
By considering values with only zeros as after the decimal point (or a maximal epsilon of 0.000001d).

See added test cases for examples.

> [OX-9546](https://scireum.myjetbrains.com/youtrack/issue/OX-9546)
> [OX-9560](https://scireum.myjetbrains.com/youtrack/issue/OX-9560)

### DIFF
--- a/src/main/java/sirius/kernel/nls/NLS.java
+++ b/src/main/java/sirius/kernel/nls/NLS.java
@@ -1229,10 +1229,12 @@ public class NLS {
                 return result;
             }
 
-            return getDecimalFormat(language).parse(value).doubleValue();
-        } catch (ParseException exception) {
-            Exceptions.ignore(exception);
-            return Double.valueOf(value);
+            try {
+                return getDecimalFormat(language).parse(value).doubleValue();
+            } catch (ParseException exception) {
+                Exceptions.ignore(exception);
+                return Double.valueOf(value);
+            }
         } catch (NumberFormatException exception) {
             throw new IllegalArgumentException(fmtr("NLS.errInvalidDecimalNumber").set("value", value).format(),
                                                exception);
--- a/src/main/java/sirius/kernel/nls/NLS.java
+++ b/src/main/java/sirius/kernel/nls/NLS.java
@@ -14,6 +14,7 @@ import sirius.kernel.async.CallContext;
 import sirius.kernel.async.ExecutionPoint;
 import sirius.kernel.commons.AdvancedDateParser;
 import sirius.kernel.commons.Amount;
+import sirius.kernel.commons.Doubles;
 import sirius.kernel.commons.Explain;
 import sirius.kernel.commons.NumberFormat;
 import sirius.kernel.commons.Strings;
@@ -1025,14 +1026,14 @@ public class NLS {
     private static <V> V parseBasicTypesFromMachineString(Class<V> clazz, String value) {
         if (Integer.class.equals(clazz) || int.class.equals(clazz)) {
             try {
-                return (V) Integer.valueOf(value);
+                return (V) parseIntegerFromMachineString(value);
             } catch (NumberFormatException e) {
                 throw new IllegalArgumentException(fmtr("NLS.errInvalidNumber").set("value", value).format(), e);
             }
         }
         if (Long.class.equals(clazz) || long.class.equals(clazz)) {
             try {
-                return (V) Long.valueOf(value);
+                return (V) parseLongFromMachineString(value);
             } catch (NumberFormatException e) {
                 throw new IllegalArgumentException(fmtr("NLS.errInvalidNumber").set("value", value).format(), e);
             }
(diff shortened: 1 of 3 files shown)


sabieber (approved):

src/main/java/sirius/kernel/nls/NLS.java
```diff
@@ -1169,6 +1190,38 @@ private static <V> V parseBasicTypesFromUserString(Class<V> clazz, String value,
         return parseDatesFromUserString(clazz, value, language);
     }
 
+    private static Integer parseIntegerFromUser(String value, String language) {
+        try {
+            Double valueAsDouble = parseDecimalNumberFromUser(value, language);
```
jakobvogel, line 1195: Probably I miss something, but isn't this exactly the code above, just nested in another catch clause? Why not invoke this?

And if it is better to copy the code, why not fire the desired exception class directly?

jakobvogel (commented):
I am puzzled... Please help me to understand the change 🤗

src/main/java/sirius/kernel/nls/NLS.java
idlira: There are basically 4 variants, slightly close but not the same.
- Integer x Long - different internal types
- FromUserString x FromMachineString - for the first we reuse the parseDoubleFromUserString, which throws a different exception as the Java's valueOf used in FromMachineString

jakobvogel (approved):
Ah, ok, now I see! 👍

src/main/java/sirius/kernel/nls/NLS.java
```diff
@@ -1025,14 +1026,14 @@ public static <V> V parseMachineString(Class<V> clazz, String value) {
...
             } catch (NumberFormatException e) {
                 throw new IllegalArgumentException(fmtr("NLS.errInvalidNumber").set("value", value).format(), e);
             }
         }
         if (Long.class.equals(clazz) || long.class.equals(clazz)) {
             try {
-                return (V) Long.valueOf(value);
+                return (V) parseLongFromMachineString(value);
             } catch (NumberFormatException e) {
                 throw new IllegalArgumentException(fmtr("NLS.errInvalidNumber").set("value", value).format(), e);
```
andyHa, line 1038: same as above but .toLongValue

src/main/java/sirius/kernel/nls/NLS.java
```diff
@@ -1025,14 +1026,14 @@ public static <V> V parseMachineString(Class<V> clazz, String value) {
     private static <V> V parseBasicTypesFromMachineString(Class<V> clazz, String value) {
         if (Integer.class.equals(clazz) || int.class.equals(clazz)) {
             try {
-                return (V) Integer.valueOf(value);
+                return (V) parseIntegerFromMachineString(value);
             } catch (NumberFormatException e) {
                 throw new IllegalArgumentException(fmtr("NLS.errInvalidNumber").set("value", value).format(), e);
```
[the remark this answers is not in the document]
andyHa: refactor this "fail is factional part into a method" and re-use 4 times

src/main/java/sirius/kernel/nls/NLS.java
```diff
@@ -1136,14 +1157,14 @@ public static <V> V parseUserString(Class<V> clazz, String value, String languag
     private static <V> V parseBasicTypesFromUserString(Class<V> clazz, String value, String language) {
         if (Integer.class.equals(clazz) || int.class.equals(clazz)) {
             try {
-                return (V) Integer.valueOf(value);
+                return (V) parseIntegerFromUser(value, language);
             } catch (NumberFormatException e) {
                 throw new IllegalArgumentException(fmtr("NLS.errInvalidNumber").set("value", value).format(), e);
```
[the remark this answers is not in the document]
andyHa: this would also add support for machine formats (see below)

src/main/java/sirius/kernel/nls/NLS.java
```diff
@@ -1136,14 +1157,14 @@ public static <V> V parseUserString(Class<V> clazz, String value, String languag
...
             }
         }
         if (Long.class.equals(clazz) || long.class.equals(clazz)) {
             try {
-                return (V) Long.valueOf(value);
+                return (V) parseLongFromUser(value, language);
             } catch (NumberFormatException e) {
                 throw new IllegalArgumentException(fmtr("NLS.errInvalidNumber").set("value", value).format(), e);
```
andyHa, line 1169: s.a.

src/main/java/sirius/kernel/nls/NLS.java
```diff
@@ -1169,17 +1190,51 @@ private static <V> V parseBasicTypesFromUserString(Class<V> clazz, String value,
...
-        } catch (ParseException exception) {
-            Exceptions.ignore(exception);
-            return Double.valueOf(value);
+            try {
+                return getDecimalFormat(language).parse(value).doubleValue();
+            } catch (ParseException exception) {
+                Exceptions.ignore(exception);
+                return Double.valueOf(value);
+            }
         } catch (NumberFormatException exception) {
```
andyHa, line 1238: i don't understand this change

sabieber (approved):

src/test/java/sirius/kernel/nls/NLSSpec.groovy
```diff
@@ -148,6 +144,26 @@ class NLSSpec extends BaseSpecification {
...
     }
 
+    def "parseUserString works for integers"() {
+        expect:
+        NLS.parseUserString(Integer.class, input) == output
+
+        where:
+        input     | output
+        "42"      | 42
+        "77,0000" | 77
```
andyHa, line 154: ensure that german and english 77.000 and 77,000 still yield 77000 and also other edge cases

andyHa (changes requested):

src/main/java/sirius/kernel/nls/NLS.java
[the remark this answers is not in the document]
idlira: Can be done, but the parsing is more complicated. "13,00" will render a BigDecimal with scale 2.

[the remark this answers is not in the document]
idlira: See new code using the common method `extractNonFactionalPart()`.
I kept the approach of using a double to detect the fractional part. A BigDecimal has no public method to obtain this information.

PR #444, opened 2023-02-28
Fixes wrong duration being displayed in the watch

mko-sci:


idlira (approved):
It fixes the problem. I wonder if the handling of nano+micro seconds should also be covered by NLS.convertDuration. Probably we don't wanna go down the road of a includeNanos or the like flag. See what others have to say.

mkeckmkeck (approved):

PR #447, opened 2023-03-15
SE-12549: Allows to extend XMLCall for Individual Response Handling (VERSION 2)

fhaScireum:
For this, the visibility for various methods and object variables was changed from private to proctected. This enables custom processing of the response in cases where the input needs intermediate processing (e.g. fixing double escaping of special characters) while still allowing the usage of the existing logging logic.

Fixes: SE-12549

### DIFF
--- a/src/main/java/sirius/kernel/xml/XMLCall.java
+++ b/src/main/java/sirius/kernel/xml/XMLCall.java
@@ -26,7 +26,7 @@ import java.net.URL;
  */
 public class XMLCall {
 
-    private Outcall outcall;
+    protected Outcall outcall;
     private NamespaceContext namespaceContext;
     private Log debugLogger = Log.get("xml");
 
@@ -146,7 +146,7 @@ public class XMLCall {
         return new XMLStructuredOutput(outcall.postFromOutput());
     }
 
-    private InputStream getInputStream() throws IOException {
+    protected InputStream getInputStream() throws IOException {
         if (debugLogger != null && debugLogger.isFINE()) {
             // log the request, even when parsing fails
             try (InputStream body = outcall.getResponse().body()) {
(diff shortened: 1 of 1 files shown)


sabieber (approved):

src/main/java/sirius/kernel/xml/XMLCall.java
```diff
@@ -26,7 +26,7 @@
  */
 public class XMLCall {
 
-    private Outcall outcall;
+    protected Outcall outcall;
```
andyHa, line 29: should not be needed...

src/main/java/sirius/kernel/xml/XMLCall.java
```diff
@@ -158,7 +158,7 @@ private InputStream getInputStream() throws IOException {
         return outcall.getResponse().body();
     }
 
-    private void logRequest(String response) throws IOException {
+    protected void logRequest(String response) throws IOException {
```
andyHa, line 161: shoudln't be needed....

src/main/java/sirius/kernel/xml/XMLCall.java
fhaScireum: We can't define the URI otherwise as the outcall requires a URI at construction and this is only done in a static method that creates and returns a simple XMLCall. 

This is an issue in the design of the class, as the default constructor seems to be pointless as no endpoint can be defined that way. We can create a protected constructor that takes a URI and content type as parameters. This will be technically breaking but the default constructor effectively doesn't work outside the class anyway.

src/main/java/sirius/kernel/xml/XMLCall.java
fhaScireum: As we want to overwrite the `getInputStream` method, we must call logRequest there manually. We can create a new protected `getResponseBody()` method, in which we do the transformation in special cases (GEVIS), but I don't see a benefit in adding an additional method that returns yet another InputStream in comparison with the approach in https://github.com/scireum/sirius-kernel/pull/446/files . We would also lose the benefit of logging the raw responses as we would need to transform them before logging.

ymo-sci (approved):

qw3ry (approved):

sabieber (approved):

idlira (approved):

andyHa (approved):

PR #452, opened 2023-05-09
Expands and tests JSON Util Class

sabieber:
The different methods now all provide a try-variant that return an empty optional when no fitting array/object node could be retrieved. The functionality of the methods without the try-prefix is changed so they fall back to an empty array/node when the try-variant didn't find a result. This replaces the old usage of withArray/withObject that actually added the array/object to the parent the retrieval was called on.

Fixes: SIRI-736

### DIFF
--- a/src/main/java/sirius/kernel/commons/Json.java
+++ b/src/main/java/sirius/kernel/commons/Json.java
@@ -9,6 +9,7 @@
 package sirius.kernel.commons;
 
 import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonPointer;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.JsonNode;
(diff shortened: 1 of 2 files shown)


andyHa (approved):

mkeckmkeck (approved):

jakobvogel (approved):

mkeckmkeck (approved):

src/main/java/sirius/kernel/commons/Json.java
```diff
@@ -267,17 +302,94 @@ public static ObjectNode getObject(ObjectNode objectNode, String fieldName) {
...
+    /**
+     * Tries to retrieve the {@link JsonNode} at the given pointer of the given {@link JsonNode}.
+     *
+     * @param jsonNode the node to retrieve the value from
+     * @param pointer  the pointer to the value to retrieve
+     * @return the value at the given pointer or an empty optional if the pointer does not exist or the node is null
+     */
+    public static Optional<JsonNode> tryGetAt(JsonNode jsonNode, JsonPointer pointer) {
+        JsonNode node = jsonNode.at(pointer);
+        if (node == null || node.isNull()) {
```
mkeckmkeck, line 390: at doesn't return null but MissingNode
```suggestion
        if (node == null || node.isNull() || node.isMissingNode()) {
```

jakobvogel (approved):

PR #464, opened 2023-06-22
Provides a central method to cleanup / normalize strings

andyHa:


sabieber (changes requested):
**Optional:** Whats missing to replace the item number normalization in OX with this is `Cleanups.UPPERCASE`.

Also I think the convention for enum names is to use the singular, so `Cleanup`, `StringCleanup` or `CleanupMode` instead of `Cleanups`.

sabieber (approved):

mkeckmkeck (approved):

idlira (approved):

PR #465, opened 2023-06-23
SIRI-823: Replaces the "..." with the single symbol "…"

oolscireum:
because the shorten(…) method removes only one symbol from the overall textlength.

Fixes: SIRI-823

### DIFF
--- a/src/main/java/sirius/kernel/commons/Strings.java
+++ b/src/main/java/sirius/kernel/commons/Strings.java
@@ -420,7 +420,7 @@ public class Strings {
         }
         String str = String.valueOf(input).trim();
         if (str.length() > length) {
-            return str.substring(0, length) + (showEllipsis ? "…" : "");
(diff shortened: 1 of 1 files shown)


idlira (changes requested):
I think we should keep the ascii chars and cut it properly since we are adding 3 chars.

andyHa (changes requested):
Also, we cannot rely on this method to correctly produce a limited string anyways (see JavaDoc) - maybe simply use Strings.limit?

andyHa:
Also, we should fix the javadoc of the method (no uppercase letter at the start of the sentence. odd linebreaks no reference to "limit" via @see

idlira (approved):

ymo-sci (approved):

fhaScireum (approved):

mkeckmkeck (approved):

andyHa (approved):

PR #470, opened 2023-07-04
Tiny docs & GC fix for Strings.cleanup

andyHa:


src/main/java/sirius/kernel/commons/Strings.java
```diff
@@ -537,7 +537,16 @@ public static String trim(Object object) {
     @Nonnull
     @SafeVarargs
     public static String cleanup(@Nullable String inputString, @Nonnull UnaryOperator<String>... cleanups) {
-        return cleanup(inputString, Arrays.asList(cleanups));
+        if (Strings.isEmpty(inputString)) {
+            return "";
```
idlira, line 541: Not sure if in this case the raw input should be returned. Thinking about things like JSON responses, might be desired that `null` stays `null` instead of being converted to an empty string.

andyHa: many cleanups being replaced by this, had this API contract to safely return a non-null string

sabieber (approved):

idlira (approved):

PR #471, opened 2023-07-09
Permits specifying a formatter in NLS.get(count)

andyHa:


src/main/java/sirius/kernel/nls/NLS.java
```diff
@@ -313,7 +346,7 @@ public static String get(@Nonnull String property, int numeric, @Nullable String
             if (versions.length == 3) {
                 return versions[0].trim();
             } else {
-                return Formatter.create(versions[1].trim()).set("count", 0).format();
+                return Formatter.create(versions[1].trim()).set("count", formatter.apply(0)).format();
             }
         } else if (numeric == 1) {
```
jakobvogel, line 351: Why do we format / support the "count" attribute for numeric 0, but not 1?

jakobvogel (approved):
Test case?

sabieber (approved):

src/main/java/sirius/kernel/nls/NLS.java
andyHa: Because this is usually a string from the NLS value e.g. "one" instead of "1"

andyHa: We support 2-part patterns: One | Many -> in this case we format the 0 case like many or 3-part patterns: Zero | One | Many

PR #476, opened 2023-07-20
Adjusts regex to only match valid html tags starting with letters 

sabieber:
Fixes: [OX-10290](https://scireum.myjetbrains.com/youtrack/issue/OX-10290)

### DIFF
--- a/src/main/java/sirius/kernel/commons/Strings.java
+++ b/src/main/java/sirius/kernel/commons/Strings.java
@@ -575,7 +575,7 @@ public class Strings {
         return value;
     }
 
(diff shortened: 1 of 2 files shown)


idlira (approved):

fhaScireum (commented):
Why is the String public? The only way to use it outside the class is in a pattern compiled form, in which case it would make more sense to directly publicize the Pattern instead.

sabieber:
> it outside the class is in a pattern compiled form, in which case it would make more sense to directly publicize the Pattern instead.

In the other larger PR from Christian (which will be overhauled after the HF) we reuse the string for building multiple patterns. But you are correct, for now I can expose the pattern and use that in OX.

fhaScireum (approved):

PR #479, opened 2023-08-01
Code Point Trie and Emoji Support

jakobvogel:
- Introduces _trie_ using code points (instead of characters), sharing a base class with the previously existing char-based trie.
- Introduces helper for working with emojis.


src/main/java/sirius/kernel/commons/Emojis.java
```diff
@@ -0,0 +1,4838 @@
...
+                if (!iterator.resetWith(sequence[index])) {
+                    return false;
+                }
+            }
+        }
+
+        return iterator.isCompleted();
+    }
+
+    public static int countEmojis(String string) {
```
jakobvogel, line 4819: Documentation follows in a further push. Sorry.

src/main/java/sirius/kernel/commons/BaseTrie.java
```diff
@@ -0,0 +1,413 @@
...
+ * A map like data structure which associates strings (char sequences) to values.
+ * <p>
+ * A trie is a highly efficient data structure for iterating through a string and retrieving a previously stored
+ * value. Checking containment or retrieving a value has guaranteed O(n) runtime, where n is the length of the
+ * processed string, independent of the size of the trie.
+ * <p>
+ * An Example: If we have a list of stop words: "one", "two", "three" and want to detect if these occur in a
+ * given text, we can do the following:
+ * <pre>
+ * {@code
```
sabieber, line 32: **Optional:** Maybe use the new `@snippet`

mko-sci (approved):

src/main/java/sirius/kernel/commons/Emojis.java
```diff
@@ -0,0 +1,4838 @@
...
+package sirius.kernel.commons;
+
+/**
+ * Provides helper methods to work with emojis.
+ */
+public class Emojis {
+
+    protected static final CodePointTrie<Boolean> emojiTrie = CodePointTrie.create();
+
+    static {
```
sabieber, line 18: **Question:** Would it make sense to extract the list of emojis into a file or config? Or would this just introduce more complexity?

sabieber (approved):

src/main/java/sirius/kernel/commons/Emojis.java
jakobvogel: That's a good question… I currently tend against it, as it would require additional facilities, and thus increase complexity. To discuss, these would be the options:

- We could add a parser for the [original file](https://unicode.org/Public/emoji/latest/emoji-test.txt), making upgrades simpler but requiring some specialisation for parsing the code points.
- We could use a file format of our own, requiring an conversion when upgrading the list.
- We can leave it in the code, also requiring conversion during upgrades, but at least saving the extra effort. (I have such a script and will provide it separately.)

Additionally, the loading times on my machine are quick enough, I currently don't see a need for lazy initialisation.

So, putting everything together, I would stick to the proposed solution for now, but i will happily change it if required.

sabieber (approved):

PR #495, opened 2023-09-28
Strings - Adds a method which counts the number of newlines in the given string

mko-sci:
Fixes: OX-10394

### DIFF
--- a/src/main/java/sirius/kernel/commons/Strings.java
+++ b/src/main/java/sirius/kernel/commons/Strings.java
@@ -393,6 +393,19 @@ public class Strings {
     }
 
     /**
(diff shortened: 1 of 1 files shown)


mkeckmkeck (commented):
Do we really only need \n ? If not, maybe use java.lang.String#lines

mko-sci:
> Do we really only need \n ? If not, maybe use java.lang.String#lines

maybe you are right, maybe its okay that we also accept
* a line feed character (“\n”)
* a carriage return character (“\r”)
* a carriage return followed immediately by a line feed (“\r\n”)
as "new line" separators...

PR #502, opened 2023-10-18
Detects entities in texts to decide if its xml

bkositza:
Texts that do not contain xml tags, but entities, should also be recognized as xml.

This facilitates the correct display of texts.

### DIFF
--- a/src/main/java/sirius/kernel/commons/Strings.java
+++ b/src/main/java/sirius/kernel/commons/Strings.java
@@ -576,7 +576,7 @@ public class Strings {
     }
 
     protected static final String REGEX_DETECT_XML_TAGS = "</?[a-zA-Z][^>]*>";
-    protected static final String REGEX_DETECT_ENTITIES = "&[a-zA-Z]+[0-9]{0,2};";
+    protected static final String REGEX_DETECT_ENTITIES = "&#?[a-zA-Z0-9]*;";
     /**
      * Defines a pattern (regular expression) to detect XML tags and entities.
      */
--- a/src/main/java/sirius/kernel/commons/Strings.java
+++ b/src/main/java/sirius/kernel/commons/Strings.java
@@ -575,17 +575,19 @@ public class Strings {
         return value;
     }
 
-    protected static final String REGEX_DETECT_XML = "</?[a-zA-Z][^>]*>";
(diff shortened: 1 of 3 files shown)


jakobvogel (approved):

fhaScireum (approved):

src/main/java/sirius/kernel/commons/Strings.java
```diff
@@ -575,17 +575,19 @@ public static String cleanup(@Nullable String inputString, @Nonnull Iterable<Una
         return value;
     }
 
-    protected static final String REGEX_DETECT_XML = "</?[a-zA-Z][^>]*>";
+    protected static final String REGEX_DETECT_XML_TAGS = "</?[a-zA-Z][^>]*>";
+    protected static final String REGEX_DETECT_ENTITIES = "&[a-zA-Z]+[0-9]{0,2};";
```
idlira, line 579: I'm not sure you cover all kinds of escape chars here:
`&#38;`, `&#x00F7;` are some samples of it...

idlira (changes requested):

mko-sci (approved):

src/main/java/sirius/kernel/commons/Strings.java
bkositza: Good point :)

src/main/java/sirius/kernel/commons/Strings.java
```diff
@@ -575,17 +575,19 @@ public static String cleanup(@Nullable String inputString, @Nonnull Iterable<Una
         return value;
     }
 
-    protected static final String REGEX_DETECT_XML = "</?[a-zA-Z][^>]*>";
+    protected static final String REGEX_DETECT_XML_TAGS = "</?[a-zA-Z][^>]*>";
+    protected static final String REGEX_DETECT_ENTITIES = "&#?[a-zA-Z0-9]*;";
```
idlira, line 579:
```suggestion
    protected static final String REGEX_DETECT_ENTITIES = "&#?[a-zA-Z0-9]+;";
```

idlira: Would just enforce that something exists between `&` and `;`. Just in case someone uses the char combination `&;` for whatever reason and we detect it as XML.

idlira (approved):

PR #505, opened 2023-10-25
Configurable Outcall blacklisting

oolscireum:
Add a blacklistid to finely adjust the Outcalls blacklisting.
This is needed for proxying servers that redirects connections to different sub servers based on configurated ids.
If one of those servers has a timeout the whole host is blocked even when other sub servers doesn't have problems.

Fixes: [SIRI-893](https://scireum.myjetbrains.com/youtrack/issue/SIRI-893)

### DIFF
--- a/src/main/java/sirius/kernel/xml/Outcall.java
+++ b/src/main/java/sirius/kernel/xml/Outcall.java
@@ -116,7 +116,7 @@ public class Outcall {
     /**
      * If the {@link #timeoutBlacklist} contains more than the given number of entries, we remove all expired ones
      * manually. These might be hosts which are only connected sporadically and had a hiccup. Everything else will be
-     * kept clean in {@link #checkTimeoutBlacklist(URI)}.
+     * kept clean in {@link #checkTimeoutBlacklist()}.
      */
     private static final int TIMEOUT_BLACKLIST_HIGH_WATERMARK = 100;
 
@@ -133,7 +133,7 @@ public class Outcall {
 
     private HttpClient client;
     private HttpRequest request;
-    private final String blacklistId;
+    private String blacklistId;
     private final HttpClient.Builder clientBuilder;
     private final HttpRequest.Builder requestBuilder;
     private HttpResponse<InputStream> response;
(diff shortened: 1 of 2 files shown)


mko-sci (approved):

mkeckmkeck (commented):
Wouldn't a wither be more appropriate to alter default settings here? Also the XMLCall cosntructor isn't accessible

src/main/java/sirius/kernel/xml/XMLCall.java
```diff
@@ -45,6 +45,19 @@ protected XMLCall(URI uri, String contentType) throws IOException {
...
 
+    /**
+     * Creates a new XMLCall for the given URI and Content-Type.
+     *
+     * @param uri         the target URI to call
+     * @param blacklistId the id for the blacklist to use.
+     * @param contentType the value of the Content-Type property for requests
+     * @throws IOException in case the URI is blacklisted
+     */
+    protected XMLCall(URI uri, String blacklistId, String contentType) throws IOException {
```
fhaScireum, line 56: Who's calling this protected constructor? I think a simple Setter in Outcall would be easiest and clearest as there is no reason why the blacklist id has to be specified when constructing the object.

fhaScireum (changes requested):

src/main/java/sirius/kernel/xml/XMLCall.java
oolscireum: The protected Constructor is overwritten with a public one in my case. I forgott to implement the methods like the XMLCall.to(...) method.
The check if a call is blacklisted happens in the Outcall constructor. A simple setter in the Outcall class would check the hostname first and then use the blacklistid later in the call.

fhaScireum (changes requested):

mko-sci (approved):

mkeckmkeck (approved):

src/main/java/sirius/kernel/xml/Outcall.java
```diff
@@ -823,4 +810,14 @@ public static Duration getDefaultConnectTimeout() {
...
         return defaultReadTimeout;
     }
+
+    /**
+     * Sets a custom blacklist id.
+     * The default blacklist id is the host of the uri.
+     *
+     * @param blacklistId The custom blacklist id to set.
+     */
+    public void setBlacklistId(@Nonnull String blacklistId) {
```
idlira, line 821: I would prefer a builder pattern, so you can simply use `new Outcall(...).withBlacklistId(...)`, avoiding the need for extra constructors.

idlira (changes requested):

src/main/java/sirius/kernel/xml/Outcall.java
oolscireum: The setter is needed for things like the XMLCall where the Outcall is created in the constructor and cannot be replaced and can only be accessed via a getter. The constructor with the blacklistid is removed and not needed anymore.

idlira (approved):

PR #508, opened 2023-11-29
Fixes Amount in Json problems with regards to precision

mkeckmkeck:
We now can read and write Amounts without changing precision.

### DIFF
--- a/src/main/java/sirius/kernel/commons/Amount.java
+++ b/src/main/java/sirius/kernel/commons/Amount.java
@@ -243,9 +243,24 @@ public class Amount extends Number implements Comparable<Amount>, Serializable {
      * @return the internally used <tt>BigDecimal</tt>
      */
     @Nullable
-    @JsonValue
     public BigDecimal getAmount() {
         return value;
+    }
+
+    /**
+     * Unwraps the internally used <tt>BigDecimal</tt> with rounding like in {@link #toMachineString()} applied.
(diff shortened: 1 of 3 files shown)


mko-sci (approved):

src/main/java/sirius/kernel/commons/Amount.java
```diff
@@ -243,11 +243,26 @@ public static Amount of(@Nullable Double amount) {
...
+     *
+     * @return the internally used <tt>BigDecimal</tt> with rounding applied
+     */
+    @Nullable
+    @JsonValue
+    private BigDecimal getRoundedAmount() {
+        if (rounded) {
+            return getAmount();
+        } else {
+            return round(2, RoundingMode.HALF_UP).getAmount();
```
jakobvogel, line 262: Hmm, if I understand this correctly, we will not be able to write more than 2 decimal digits to JSON? Or could I manually round to, say, 3 digits and have that in JSON?

mkeckmkeck: Yes, this is alligning the behavior with toMachineString. When no specific round is set, the default precision is 2. 
See testcase on how Amount with custom precision is used.

jakobvogel (approved):

idlira (approved):

ymo-sci (approved):

PR #511, opened 2023-12-08
Reduces logging for blacklisted hosts

bkositza:
We log every attempt to call a blacklisted host again. This PR limits this logging to one message per host so that the logs are not flooded with frequent calls.

### DIFF
--- a/src/main/java/sirius/kernel/xml/Outcall.java
+++ b/src/main/java/sirius/kernel/xml/Outcall.java
@@ -15,6 +15,7 @@ import sirius.kernel.commons.Explain;
 import sirius.kernel.commons.Monoflop;
 import sirius.kernel.commons.Streams;
 import sirius.kernel.commons.Strings;
+import sirius.kernel.commons.Tuple;
 import sirius.kernel.commons.Watch;
 import sirius.kernel.di.std.ConfigValue;
 import sirius.kernel.health.Average;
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/kernel/xml/Outcall.java
```diff
@@ -500,15 +504,25 @@ private void checkTimeoutBlacklist() throws IOException {
             return;
         }
 
-        Long timeout = timeoutBlacklist.get(blacklistId);
-        if (timeout != null) {
-            if (timeout > System.currentTimeMillis()) {
+        Tuple<Long, Boolean> blacklistedHostInformation = timeoutBlacklist.get(blacklistId);
+        if(blacklistedHostInformation == null) {
```
idlira, line 508: Code Formatting?
```suggestion
        if (blacklistedHostInformation == null) {
```

src/main/java/sirius/kernel/xml/Outcall.java
```diff
@@ -518,11 +532,11 @@ private void addToTimeoutBlacklist() {
         }
 
         long now = System.currentTimeMillis();
-        timeoutBlacklist.put(blacklistId, now + connectTimeoutBlacklistDuration.toMillis());
+        timeoutBlacklist.put(blacklistId, new Tuple<>(now + connectTimeoutBlacklistDuration.toMillis(), false));
```
idlira, line 535: Use `Tuple.create`

src/main/java/sirius/kernel/xml/Outcall.java
```diff
@@ -110,8 +111,11 @@ public class Outcall {
      * <p>
      * These hosts are blacklisted for a short amount of time ({@link #connectTimeoutBlacklistDuration}) to prevent
      * cascading failures.
+     * <p>
+     * We log a warning for each host which is blacklisted and is called again. To make sure this doesn't spam the logs,
+     * we only log once and then set the second value of the tuple to <tt>true</tt> to prevent further log messages.
      */
-    private static final Map<String, Long> timeoutBlacklist = new ConcurrentHashMap<>();
+    private static final Map<String, Tuple<Long, Boolean>> timeoutBlacklist = new ConcurrentHashMap<>();
```
idlira, line 118: **Optional**: a **record** would be cleaner than a Tuple and could encapsulate some helper methods to substitute the null checks done below.

mko-sci (approved):

oolscireum (approved):

src/main/java/sirius/kernel/xml/Outcall.java
bkositza: You're right, a record would be cleaner and easier to understand. I also tried a record at first, but I have to admit that - for this small task - a tuple might be enough. 

Correct me if I'm wrong, but a record only transports data that is final. Therefore, we could not call a setter, but would always have to recreate the record, or?

PR #514, opened 2024-01-04
Expands the sub-context helpers of the CallContext

sabieber:


src/main/java/sirius/kernel/async/CallContext.java
```diff
@@ -353,10 +353,46 @@ public <C extends SubContext> C get(Class<C> contextType) {
...
+     *
+     * @param contextType the type of the sub-context to be returned
+     * @param <C>         the type of the sub-context
+     * @return an instance of the given type or <tt>null</tt> if no instance was available
+     */
+    @Nullable
+    @SuppressWarnings("unchecked")
+    public <C extends SubContext> C tryGetSubContext(@Nonnull Class<C> contextType) {
+        return (C) subContexts.get(contextType);
+    }
```
jakobvogel, line 405: The function name originally made me believe the return type would be an `Optional` 😅

jakobvogel (approved):

src/main/java/sirius/kernel/async/CallContext.java
idlira: Yes, I prefer if an Optional comes out if the method is called `try***`
Otherwise rename the other one to getRequredSubContext or the like...

mkeckmkeck (approved):

idlira (approved):

jakobvogel (approved):

PR #516, opened 2024-02-11
Allows defining a custom Content-Type header for SOAP requests.

fhaScireum:
By default, the underlying XMLCall sets the header to "text/xml" if left underfined, but some servers may require a different header (e.g. "text/xml; charset=utf-8").

Fixes: SE-13428

### DIFF
--- a/src/main/java/sirius/kernel/xml/SOAPClient.java
+++ b/src/main/java/sirius/kernel/xml/SOAPClient.java
@@ -98,11 +98,13 @@ public class SOAPClient {
      */
     public static final String SOAP_TIMEOUT_CONFIG_KEY = "soap";
 
+    private static final String DEFAULT_CONTENT_TYPE_HEADER = "text/xml";
+
     private final URL endpoint;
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/kernel/xml/SOAPClient.java
```diff
@@ -169,6 +170,18 @@ public SOAPClient withActionPrefix(@Nonnull String prefix) {
         return this;
     }
 
+    /**
+     * Defines a custom value for the "Content-Type" HTTP header to use for requests. By default, the header is set
+     * to "text/xml".
```
sabieber, line 175: **Optional:** Why not define `text/xml` as a constant like `DEFAULT_CONTENT_TYPE` and link it here? Then the info is less likely to become outdated when the default is somehow changed.

sabieber (approved):

mko-sci (approved):

ymo-sci (approved):

PR #517, opened 2024-03-21
Allow 1 and 0 for boolean values in Value#asBoolean

ymo-sci:
in https://www.w3.org/TR/2004/REC-xmlschema-2-20041028/#boolean 1 and 0 are also allowed as valid boolean values. In Some places we read from XML and have StructuredNode.queryValue("VALUE_WITH_0_OR_1").asBoolean(false) which should also return true if the value is 1

BREAKING: Value#coerce(Boolean.class, true) now return true as expected and no longer false if the value is empty

- fixes: SE-13483

### DIFF
--- a/src/main/java/sirius/kernel/commons/Value.java
+++ b/src/main/java/sirius/kernel/commons/Value.java
@@ -534,7 +534,7 @@ public class Value {
                 return (T) data;
             }
             String stringValue = String.valueOf(data).trim();
-            return (T) fastPathBoolean(stringValue).orElseGet(() -> NLS.parseMachineString(Boolean.class, stringValue));
+            return (T) parseWithoutNLS(stringValue).orElseGet(() -> NLS.parseMachineString(Boolean.class, stringValue));
         }
         if (data == null) {
             return defaultValue;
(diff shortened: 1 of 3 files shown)


src/main/java/sirius/kernel/commons/Value.java
```diff
@@ -813,16 +813,21 @@ public boolean asBoolean(boolean defaultValue) {
...
-        // fast-track for common cases without the need to involve NLS framework
-        if ("true".equalsIgnoreCase(String.valueOf(data))) {
-            return true;
+        String stringValue = String.valueOf(data);
+        return fastPathBoolean(stringValue).orElseGet(() -> {
+            return Objects.requireNonNullElse(NLS.parseUserString(Boolean.class, stringValue.trim()), defaultValue);
+        });
+    }
+
+    // fast-track for common boolean cases without the need to involve NLS framework
```
idlira, line 823: Use real JavaDoc to describe a method (even private).

src/main/java/sirius/kernel/commons/Value.java
```diff
@@ -813,16 +813,21 @@ public boolean asBoolean(boolean defaultValue) {
...
+    private Optional<Boolean> fastPathBoolean(String value) {
```
idlira, line 824: raname to `parseWithoutNLS`

idlira (changes requested):

mkeckmkeck (approved):

src/main/java/sirius/kernel/commons/Value.java
```diff
@@ -802,7 +803,7 @@ public String asSmartRoundedString() {
      *
      * @param defaultValue the value to be used if the wrapped value cannot be converted to a boolean.
      * @return <tt>true</tt> if the wrapped value is <tt>true</tt>
-     * or if the string representation of it is {@code "true"}. Returns <tt>false</tt> otherwise,
+     * or if the string representation of it is {@code "true"} or {@code "1!"}. Returns <tt>false</tt> otherwise,
```
sabieber, line 806:
```suggestion
     * or if the string representation of it is {@code "true"} or {@code "1"}. Returns <tt>false</tt> otherwise,
```
Typo, right?

sabieber (approved):

idlira (approved):

andyHa (approved):

PR #526, opened 2024-05-16
 Extends the limit method in order to support preserving chars at the end of the input

mko-sci:
- This is helpful when truncating error messages as often the helpful information is at the end of the String.
- Also adds an option to add a marker/signal which states that the input has been truncated to increase the visibility for a user reading the truncated String.
- Also handles negativ `length`

Bildschirmfoto 2024-05-16 um 14 56 04

### Additional Notes

- This PR fixes or works on following ticket(s): [SIRI-960](https://scireum.myjetbrains.com/youtrack/issue/SIRI-960)

### Checklist

### DIFF
--- a/src/main/java/sirius/kernel/commons/Strings.java
+++ b/src/main/java/sirius/kernel/commons/Strings.java
@@ -395,6 +395,42 @@ public class Strings {
     }
 
     /**
+     * Truncates the given string while preserving characters at the end. Also adds a truncated signal in the middle to
+     * indicate that the string was truncated.
+     * <p>
+     * Note:
+     * The truncated string consists of three parts:
+     * <ul>
+     *     <li>
+     *         start - this is the first part of the string for which the algorithm tries to calculate an effective
+     *         length. This is done by subtracting the length of the truncated signal (13) and the characters to
+     *         preserve at the end from the given <tt>length</tt>. If for some reason the calculated effective length is
+     *         negative, the original provided length is taken. This means that the total length of the returned string
+     *         could be length + 13 + charactersToPreserveAtTheEnd. This is only the case if either the length is very
+     *         small or the charactersToPreserveAtTheEnd is ridiculously high.
+     *     </li>
+     *     <li>
+     *         middle - this is the truncated signal which is added in the middle of the truncated string. It consists
+     *         of 13 characters in total (<tt>…[truncated]…</tt>).
+     *     </li>
+     *     <li>
+     *         end - this is the last part of the original string and contains the characters which should be
+     *         preserved at the end of the input.
+     *     </li>
+     * </ul>
+     *
+     * @param input                        the object which string representation should be limited to the given length
+     * @param length                       the max. number of characters to return
+     * @param charactersToPreserveAtTheEnd the number of characters to preserve at the end of the string
+     * @return the truncated string with a signal in the middle and characters preserved at the end, or the full string
+     * if the <tt>input</tt> is shorter than the given length + charactersToPreserveAtTheEnd. However,
+     * if input is <tt>null</tt>, "" is returned.
+     */
+    public static String truncateMiddle(@Nullable Object input, int length, int charactersToPreserveAtTheEnd) {
+        return limit(input, length, true, true, charactersToPreserveAtTheEnd);
+    }
+
+    /**
      * Limits the length of the given string to the given length.
      *
      * @param input                        the object which string representation should be limited to the given length
@@ -406,11 +442,11 @@ public class Strings {
      * than <tt>length</tt>, the full value is returned. If input is <tt>null</tt>, "" is returned.
      * If <tt>charactersToPreserveAtTheEnd</tt> is negativ it is omitted.
      */
-    public static String limit(@Nullable Object input,
-                               int length,
-                               boolean showEllipsis,
-                               boolean showTruncated,
-                               int charactersToPreserveAtTheEnd) {
+    private static String limit(@Nullable Object input,
+                                int length,
+                                boolean showEllipsis,
+                                boolean showTruncated,
+                                int charactersToPreserveAtTheEnd) {
         if (isEmpty(input)) {
             return "";
         }
@@ -424,15 +460,15 @@ public class Strings {
             charactersToPreserveAtTheEnd = 0;
         }
 
-        String signal = generateTruncatedSignal(showEllipsis, showTruncated, charactersToPreserveAtTheEnd > 0);
-        int effectiveLength = length - signal.length() - charactersToPreserveAtTheEnd;
+        String middle = generateTruncatedSignal(showEllipsis, showTruncated, charactersToPreserveAtTheEnd > 0);
+        int effectiveLength = length - middle.length() - charactersToPreserveAtTheEnd;
         if (effectiveLength < 0) {
             effectiveLength = length;
         }
 
-        String truncatedStr = str.substring(0, effectiveLength).trim();
+        String start = str.substring(0, effectiveLength).trim();
         String end = str.substring(str.length() - charactersToPreserveAtTheEnd).trim();
-        return truncatedStr + signal + end;
+        return start + middle + end;
     }
 
     private static String generateTruncatedSignal(boolean showEllipsis,
--- a/src/main/java/sirius/kernel/commons/Strings.java
+++ b/src/main/java/sirius/kernel/commons/Strings.java
@@ -452,8 +452,12 @@ public class Strings {
         }
 
         String str = String.valueOf(input).trim();
-        if (str.length() <= length + charactersToPreserveAtTheEnd || length < 0) {
+        if (str.length() <= length + charactersToPreserveAtTheEnd) {
             return str;
+        }
+
+        if (length < 0) {
+            length = 0;
         }
 
         if (charactersToPreserveAtTheEnd < 0) {
--- a/src/main/java/sirius/kernel/commons/Strings.java
+++ b/src/main/java/sirius/kernel/commons/Strings.java
@@ -39,6 +39,16 @@ import java.util.stream.Collectors;
 public class Strings {
 
     /**
+     * Contains the characters which should be used in order to depict an ellipsis.
+     */
+    private static final String ELLIPSIS = "…";
+
+    /**
+     * Contains the marker which should be used to depict that the string has been truncated.
+     */
+    private static final String TRUNCATED_MARKER = "[truncated]";
+
+    /**
      * Contains all characters which can safely be used for codes without too much confusion (e.g. 0 vs O are
      * excluded).
      */
@@ -410,22 +420,19 @@ public class Strings {
             return str;
         }
 
-        final String ellipsis = "…";
-        final String truncated = "[truncated]";
-
         if (charactersToPreserveAtTheEnd < 0) {
             charactersToPreserveAtTheEnd = 0;
         }
 
         StringBuilder signal = new StringBuilder();
         if (showEllipsis) {
-            signal.append(ellipsis);
+            signal.append(ELLIPSIS);
         }
         if (showTruncated) {
-            signal.append(truncated);
+            signal.append(TRUNCATED_MARKER);
         }
         if (charactersToPreserveAtTheEnd > 0 && showTruncated && showEllipsis) {
-            signal.append(ellipsis);
+            signal.append(ELLIPSIS);
         }
 
         int effectiveLength = length - signal.length() - charactersToPreserveAtTheEnd;
--- a/src/main/java/sirius/kernel/commons/Strings.java
+++ b/src/main/java/sirius/kernel/commons/Strings.java
@@ -451,17 +451,17 @@ public class Strings {
             return "";
         }
 
-        String str = String.valueOf(input).trim();
-        if (str.length() <= length + charactersToPreserveAtTheEnd) {
-            return str;
-        }
-
         if (length < 0) {
             length = 0;
         }
 
         if (charactersToPreserveAtTheEnd < 0) {
             charactersToPreserveAtTheEnd = 0;
+        }
+
+        String str = String.valueOf(input).trim();
+        if (str.length() <= length + charactersToPreserveAtTheEnd) {
+            return str;
         }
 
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/kernel/commons/Strings.java
```diff
@@ -381,15 +381,61 @@ public static String limit(@Nullable Object input, int length) {
...
+     * @param charactersToPreserveAtTheEnd the number of characters to preserve at the end of the string
+     * @return a part of the string representation of the given <tt>input</tt>. If input is shorter
+     * than <tt>length</tt>, the full value is returned. If input is <tt>null</tt>, "" is returned.
+     * If <tt>charactersToPreserveAtTheEnd</tt> is negativ it is omitted.
+     */
+    public static String limit(@Nullable Object input,
+                               int length,
+                               boolean showEllipsis,
+                               boolean showTruncated,
+                               int charactersToPreserveAtTheEnd) {
```
sabieber, line 403: Do all combinations of these parameters actually make sense? Wouldn't it be clearer if the "truncate in the middle" function would be separate instead of cramming it into the limit function?

Can be quite confusion to set them up to make exactly what you want.
If we split them we could also provide some small examples in the JavaDoc :)

src/main/java/sirius/kernel/commons/Strings.java
```diff
@@ -381,15 +381,61 @@ public static String limit(@Nullable Object input, int length) {
...
         String str = String.valueOf(input).trim();
-        if (str.length() > length) {
-            return str.substring(0, (showEllipsis ? length - 1 : length)) + (showEllipsis ? "…" : "");
-        } else {
+        if (str.length() <= length) {
             return str;
         }
+
+        final String ellipsis = "…";
+        final String truncated = "[truncated]";
```
sabieber, line 414: Why are these not constants?

src/main/java/sirius/kernel/commons/Strings.java
```diff
@@ -381,15 +381,61 @@ public static String limit(@Nullable Object input, int length) {
...
+        StringBuilder signal = new StringBuilder();
+        if (showEllipsis) {
+            signal.append(ellipsis);
+        }
+        if (showTruncated) {
+            signal.append(truncated);
+        }
+        if (charactersToPreserveAtTheEnd > 0 && showTruncated && showEllipsis) {
+            signal.append(ellipsis);
+        }
```
sabieber, line 429: So we have to set both flags to true to get the nice output from the PR screenshot? In my opinion that is not very clear from the JavaDoc

src/main/java/sirius/kernel/commons/Strings.java
```diff
@@ -381,15 +381,61 @@ public static String limit(@Nullable Object input, int length) {
...
+
+        int effectiveLength = length - signal.length() - charactersToPreserveAtTheEnd;
+        if (effectiveLength < 0) {
+            effectiveLength = length;
+        }
+
+        String truncatedStr = str.substring(0, effectiveLength).trim();
```
sabieber, line 436: **Optional:** Don't like truncated variable names 😜

src/main/java/sirius/kernel/commons/Strings.java
mko-sci: Why is it not clear? :D
If you want do have ellipsis you need to set the flag to `true`. 
If you want the marker you need to set the flag to `true`.
If you want just one....

sabieber: See your biz PR, you actually build a helper there because you dont like to configure so many parameters multiple times

mko-sci: I did it to make sure that everything is set equal since I call it multiple times there. 
Would have done the same if the method would only have 3 params

sabieber (approved):

mkeckmkeck (approved):

src/test/kotlin/sirius/kernel/commons/StringsTest.kt
```diff
@@ -332,4 +332,90 @@ class StringsTest {
...
+        assertEquals("ABCDEFG-AB…[truncated]", Strings.truncateMiddle("ABCDEFG-ABCDEFG-ABCDEFG-ABCDEFG-ABCDEFG-ABCDEFG", 10, -10))
+        // Szenario 5: input present and length is equal to input.length() and charactersToPreserveAtTheEnd is 0 or negative
+        assertEquals("ABCDEFG", Strings.truncateMiddle("ABCDEFG", 7, 0))
+        assertEquals("ABCDEFG", Strings.truncateMiddle("ABCDEFG", 7, -10))
+        // Szenario 6: input present and length is 0 and charactersToPreserveAtTheEnd is 0 or negative
+        assertEquals("…[truncated]", Strings.truncateMiddle("ABCDEFG", 0, 0))
+        assertEquals("…[truncated]", Strings.truncateMiddle("ABCDEFG", 0, -10))
+        // Szenario 7: input present and length is negative and charactersToPreserveAtTheEnd is 0 or negative
+        assertEquals("…[truncated]", Strings.truncateMiddle("ABCDEFG", -10, 0))
+        assertEquals("…[truncated]", Strings.truncateMiddle("ABCDEFG", -10, -10))
```
sabieber, line 372: In my opinion these cases are bad behaviour, as the replacement is longer than the input string

src/test/kotlin/sirius/kernel/commons/StringsTest.kt
```diff
@@ -332,4 +332,90 @@ class StringsTest {
...
+        assertEquals("ABCDEFG", Strings.truncateMiddle("ABCDEFG", 10, 8))
+        // Szenario 10: input present and length is greater than input.length() and charactersToPreserveAtTheEnd is equal to input.length()
+        assertEquals("ABCDEFG", Strings.truncateMiddle("ABCDEFG", 10, 7))
+        // Szenario 11: input present and length is smaller than input.length() and charactersToPreserveAtTheEnd is greater than length and length + charactersToPreserveAtTheEnd is smaller than input.length()
+        // Note: the ellipsis after the truncated signal is only added if there are characters to preserve at the end
+        // Note: the length of the truncated signal is only taken into account when limiting the input to length if length is greater than 13, otherwise the input will be truncated to the provided length and the truncated signal is added afterward
+        assertEquals("ABCDEFG-…[truncated]…This is the important part", Strings.truncateMiddle("ABCDEFG-ABCDEFG-This is the important part", 8, 26))
+        assertEquals("ABCDEFG-".length, 8)
+        assertEquals("This is the important part".length, 26)
```
sabieber, line 385: What does this test?

src/test/kotlin/sirius/kernel/commons/StringsTest.kt
```diff
@@ -332,4 +332,90 @@ class StringsTest {
...
+        // Szenario 9: input present and length is greater than input.length() and charactersToPreserveAtTheEnd is greater than input.length()
+        assertEquals("ABCDEFG", Strings.truncateMiddle("ABCDEFG", 10, 8))
+        // Szenario 10: input present and length is greater than input.length() and charactersToPreserveAtTheEnd is equal to input.length()
+        assertEquals("ABCDEFG", Strings.truncateMiddle("ABCDEFG", 10, 7))
+        // Szenario 11: input present and length is smaller than input.length() and charactersToPreserveAtTheEnd is greater than length and length + charactersToPreserveAtTheEnd is smaller than input.length()
+        // Note: the ellipsis after the truncated signal is only added if there are characters to preserve at the end
+        // Note: the length of the truncated signal is only taken into account when limiting the input to length if length is greater than 13, otherwise the input will be truncated to the provided length and the truncated signal is added afterward
+        assertEquals("ABCDEFG-…[truncated]…This is the important part", Strings.truncateMiddle("ABCDEFG-ABCDEFG-This is the important part", 8, 26))
```
sabieber, line 383: same as above

src/test/kotlin/sirius/kernel/commons/StringsTest.kt
```diff
@@ -332,4 +332,90 @@ class StringsTest {
...
+        // Note: in the following case it is assumed that the whole string should be preserved because charactersToPreserveAtTheEnd is greater than input.length()
+        assertEquals("ABCDEFG-ABCDEFG-This is the important part", Strings.truncateMiddle("ABCDEFG-ABCDEFG-This is the important part", 0, 260))
+        assertEquals("ABCDEFG-ABCDEFG-This is the important part", Strings.truncateMiddle("ABCDEFG-ABCDEFG-This is the important part", -10, 260))
+        // Szenario 13: input present and length is smaller than input.length() and charactersToPreserveAtTheEnd is equal to length and length + charactersToPreserveAtTheEnd is equal to input.length()
+        // Note: in the following case it is assumed that the whole string should be preserved because length + charactersToPreserveAtTheEnd is equal to input.length()
+        assertEquals("ABCDEFG-ABCDEFG-This is the important part", Strings.truncateMiddle("ABCDEFG-ABCDEFG-This is the important part", 21, 21))
+        assertEquals("ABCDEFG-ABCDEFG-This is the important part".length, 42)
```
sabieber, line 393: same as above

src/test/kotlin/sirius/kernel/commons/StringsTest.kt
```diff
@@ -332,4 +332,90 @@ class StringsTest {
...
+        // Szenario 14: input present and length is equal to input.length() and charactersToPreserveAtTheEnd is greater than length
+        assertEquals("ABCDEFG", Strings.truncateMiddle("ABCDEFG", 7, 8))
+        // Szenario 15: input present and length is equal to input.length() and charactersToPreserveAtTheEnd is greater than input.length()
+        assertEquals("ABCDEFG", Strings.truncateMiddle("ABCDEFG", 7, 8000))
+        // Szenario 16: input present and length is equal to input.length() and charactersToPreserveAtTheEnd is equal to input.length()
+        assertEquals("ABCDEFG", Strings.truncateMiddle("ABCDEFG", 7, 7))
+        // Szenario 17: input present and length is 0 and charactersToPreserveAtTheEnd is greater than length
+        assertEquals("…[truncated]…FG", Strings.truncateMiddle("ABCDEFG", 0, 2))
```
sabieber, line 401: same as above

src/test/kotlin/sirius/kernel/commons/StringsTest.kt
```diff
@@ -332,4 +332,90 @@ class StringsTest {
...
+        // Szenario 18: input present and length is 0 and charactersToPreserveAtTheEnd is greater than input.length()
+        assertEquals("ABCDEFG", Strings.truncateMiddle("ABCDEFG", 0, 10))
+        // Szenario 19: input present and length is 0 and charactersToPreserveAtTheEnd is equal to input.length()
+        assertEquals("ABCDEFG", Strings.truncateMiddle("ABCDEFG", 0, 7))
+        // Szenario 20: input present and length is negative and charactersToPreserveAtTheEnd is greater than length
+        assertEquals("…[truncated]…DEFG", Strings.truncateMiddle("ABCDEFG", -10, 4))
```
sabieber, line 407: same as above, the replacement should not be longer than the input, just keep the input in those cases

src/main/java/sirius/kernel/commons/Strings.java
```diff
@@ -381,15 +391,108 @@ public static String limit(@Nullable Object input, int length) {
...
+
+        if (charactersToPreserveAtTheEnd < 0) {
+            charactersToPreserveAtTheEnd = 0;
+        }
+
         String str = String.valueOf(input).trim();
-        if (str.length() > length) {
-            return str.substring(0, (showEllipsis ? length - 1 : length)) + (showEllipsis ? "…" : "");
-        } else {
+        if (str.length() <= length + charactersToPreserveAtTheEnd) {
```
sabieber, line 467: **Optional:** would be nice to improve the variable name `str` as this is as unhelpful as it gets

src/test/kotlin/sirius/kernel/commons/StringsTest.kt
mko-sci: the behavior is copied from the old `limit`. 
If I change it, it gets changed there as well. 
If you are oky with that I can change it.

src/test/kotlin/sirius/kernel/commons/StringsTest.kt
mko-sci: Verify that my assumption holds. Its part of the statement above

src/test/kotlin/sirius/kernel/commons/StringsTest.kt
mko-sci: yeah we can discuss about that...
What should be the alternative behavior?
You just don't get the "truncated signal" although you set it to `true`?

src/test/kotlin/sirius/kernel/commons/StringsTest.kt
mko-sci: What is the issue here?
This seems fine to me?

src/test/kotlin/sirius/kernel/commons/StringsTest.kt
mko-sci: It should show that the start part got limited to the provided lenth and the second statement shows that the end part has all chars that we requested, but I should probably extract it in params to make it more clear

sabieber: But it just tests that the Java language feature "String::length works ... thats not our responsibility

src/test/kotlin/sirius/kernel/commons/StringsTest.kt
sabieber: it doesnt test anything

src/main/java/sirius/kernel/commons/Strings.java
sabieber: Don't really like the included type in the new name, but its okay :) better than before

mko-sci:
Has been closed since we decided to separate the `truncateMiddle` logic completely from the one used in `limit`. 
A follow up PR is in the making...

PR #528, opened 2024-05-22
Introduces `truncateMiddle` which is capable of truncating a string in the middle

mko-sci:
### Additional Notes

- This PR fixes or works on following ticket(s): [SIRI-960](https://scireum.myjetbrains.com/youtrack/issue/SIRI-960)

### Checklist

### DIFF
--- a/src/test/kotlin/sirius/kernel/commons/StringsTest.kt
+++ b/src/test/kotlin/sirius/kernel/commons/StringsTest.kt
@@ -319,4 +319,25 @@ class StringsTest {
         assertEquals("A", Strings.limit("A", 1, false))
     }
 
+    @Test
+    fun truncateMiddle() {
+        assertEquals("", Strings.truncateMiddle(null, 4, 4))
+        assertEquals("", Strings.truncateMiddle(null, -4, 4))
+        assertEquals("", Strings.truncateMiddle(null, 4, -4))
+        assertEquals("", Strings.truncateMiddle("", 4, 4))
+        assertEquals("", Strings.truncateMiddle("", -4, 4))
+        assertEquals("", Strings.truncateMiddle("", 4, -4))
+        assertEquals("", Strings.truncateMiddle("1234-6789", -4, 4))
+        assertEquals("", Strings.truncateMiddle("1234-6789", 4, -4))
+        assertEquals("", Strings.truncateMiddle("1234-6789", -4, -4))
+        assertEquals("", Strings.truncateMiddle("1234-6789", 0, 0))
+        assertEquals("1234-6789", Strings.truncateMiddle("1234-6789", 40, 0))
+        assertEquals("1234-6789", Strings.truncateMiddle("1234-6789", 0, 40))
+        assertEquals("1234-6789", Strings.truncateMiddle("1234-6789", 5, 4))
+        assertEquals("1234-6789", Strings.truncateMiddle("1234-6789", 4, 5))
+        assertEquals("1234…[truncated]…6789", Strings.truncateMiddle("1234-6789", 4, 4))
+        assertEquals("…[truncated]…6789", Strings.truncateMiddle("1234-6789", 0, 4))
+        assertEquals("1234…[truncated]…", Strings.truncateMiddle("1234-6789", 4, 0))
+        assertEquals("1…[truncated]…9", Strings.truncateMiddle("1234-6789", 1, 1))
+    }
 }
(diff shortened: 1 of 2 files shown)


scireum-mbo (approved):

idlira (approved):

src/test/kotlin/sirius/kernel/commons/StringsTest.kt
```diff
@@ -319,4 +319,25 @@ class StringsTest {
...
+        assertEquals("", Strings.truncateMiddle("", 4, -4))
+        assertEquals("", Strings.truncateMiddle("1234-6789", -4, 4))
+        assertEquals("", Strings.truncateMiddle("1234-6789", 4, -4))
+        assertEquals("", Strings.truncateMiddle("1234-6789", -4, -4))
+        assertEquals("", Strings.truncateMiddle("1234-6789", 0, 0))
+        assertEquals("1234-6789", Strings.truncateMiddle("1234-6789", 40, 0))
+        assertEquals("1234-6789", Strings.truncateMiddle("1234-6789", 0, 40))
+        assertEquals("1234-6789", Strings.truncateMiddle("1234-6789", 5, 4))
+        assertEquals("1234-6789", Strings.truncateMiddle("1234-6789", 4, 5))
+        assertEquals("1234…[truncated]…6789", Strings.truncateMiddle("1234-6789", 4, 4))
```
sabieber, line 338: **Question:** Here the result gets longer as the input again?? whats the usecase of this method if it makes a text longer in some cases?

My suggestion: Always return the original String when (preserveStart + signal.length + preserveEnd) is greater or equal to the input.length

mko-sci: *Answer:* 
- the use-case is to truncate Strings with a signal showing that something has been truncated. For example: lets say I dont care for the first part of the String, which is why I set `charsToPreserveAtStart` to `0` and the other parameter to `15` because I only want to see the last 15 chars... Then it would be helpful to see that something got `truncated`. Since I don't know how long the input will be, at least my output will be "stable" as its always signal + `charsToPreserveAtEnd` in this case.

But I can change it accordingly if you like

sabieber: But when I want to truncate a string its because I want to restrict it to a certain length, so making it longer by "truncating" it defeats the purpose in my opinion.

> Then it would be helpful to see that something got `truncated`
Thats not what I am critiquing, I am critiquing that we truncate when we don't gain a benefit from it, but instead a negative (as in lost information and sometimes even a longer string than before)

mko-sci: IMHO: Changing it to your suggestion makes the method less intuitive because then I need to know that the string has to be a minimum length of 14 chars before truncating takes place, also I need to make sure that the sum of `charsToPreserveFromStart`, `charsToPreserveFromEnd`, `13` makes sense, otherwise I get the input string back

mko-sci: Discussion continued in google-meet...

sabieber (approved):

src/test/kotlin/sirius/kernel/commons/StringsTest.kt
```diff
@@ -319,4 +319,30 @@ class StringsTest {
...
+        assertEquals("", Strings.truncateMiddle("1234-6789", 0, 0))
+        assertEquals("1234-6789", Strings.truncateMiddle("1234-6789", 40, 0))
+        assertEquals("1234-6789", Strings.truncateMiddle("1234-6789", 0, 40))
+        assertEquals("1234-6789", Strings.truncateMiddle("1234-6789", 5, 4))
+        assertEquals("1234-6789", Strings.truncateMiddle("1234-6789", 4, 5))
+        assertEquals("1234-6789", Strings.truncateMiddle("1234-6789", 4, 4))
+        assertEquals("1234-6789", Strings.truncateMiddle("1234-6789", 0, 4))
+        assertEquals("1234-6789", Strings.truncateMiddle("1234-6789", 4, 0))
+        assertEquals("1234-6789", Strings.truncateMiddle("1234-6789", 1, 1))
+        assertEquals("1234…[truncated]…6789", Strings.truncateMiddle("123456789-123456789-123456789", 4, 4))
```
ymo-sci, line 342: truncated noch mit NLS... ?

ymo-sci (approved):

src/test/kotlin/sirius/kernel/commons/StringsTest.kt
mko-sci: hmmm

mko-sci: Not a huge fan. I think I will replace `truncated` with an ellipsis, than it's just symbols and should not need a translation. 
=> 
`…[truncated]…` will be changed to `…[…]…`

idlira (approved):

sabieber (approved):

jakobvogel (approved):

PR #533, opened 2024-08-08
Adds date comparison methods

idlira:
### Description

Unfortunately Java's standard methods do not provide `afterOrEqual` or `beforeOrEqual`, nor a range check.

Note that we are leaving null-handling for the caller to decide, as it strongly depends on usage: am empty date might represent an infinite date in the past or in the future, or represent an exception.

### Checklist

### DIFF
--- a/src/main/java/sirius/kernel/commons/Dates.java
+++ b/src/main/java/sirius/kernel/commons/Dates.java
@@ -8,6 +8,9 @@
 
 package sirius.kernel.commons;
 
+import com.google.common.collect.Range;
+
+import javax.annotation.Nonnull;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


jakobvogel (approved):
Reformat and new methods would have been better in two commits. And for the relatively simple methods, is it really worth to use a helper class? (Referring to the use of `Range`)

idlira:
> Reformat and new methods would have been better in two commits. And for the relatively simple methods, is it really worth to use a helper class? (Referring to the use of `Range`)

Range comes from Google, and we hope to have a better alternative one day. And since we already provide a Dates class, it kinda suits for it.

PS: sorry for the reformat.

jakobvogel:
> > Reformat and new methods would have been better in two commits. And for the relatively simple methods, is it really worth to use a helper class? (Referring to the use of `Range`)
> 
> Range comes from Google, and we hope to have a better alternative one day. And since we already provide a Dates class, it kinda suits for it.

Hmm, for instance, `Range.atMost(referenceDate).contains(dateToCheck)` should be the same as `referenceDate.isAfter(dateToCheck)` if I am not wrong? The `Range` methods do not seem to support `null` values and also throw exceptions in such cases. And they allocate an internal object that is likely not required when just comparing the values. Correct me if I am wrong, but I do not see any particular advantage of using `Range`?

Extending the `Dates` class is fine, I was only wondering about the use of `Range`.

jakobvogel (approved):

mkeckmkeck (approved):

fhaScireum (approved):

PR #544, opened 2024-12-10
Files: Implements methods isHidden and isMetadata to check if a path is either hidden or a type of metadata file

scireum-mbo:
### Additional Notes

- This PR fixes or works on following ticket(s): [SIRI-1028](https://scireum.myjetbrains.com/youtrack/issue/SIRI-1028)
- This PR is related to PR: https://github.com/scireum/sirius-biz/pull/2059

### Checklist

- [ ] Patch Tasks: Is local execution of Patch Tasks necessary? If so, please also mark the PR with the tag.

### DIFF
--- a/src/main/java/sirius/kernel/commons/Files.java
+++ b/src/main/java/sirius/kernel/commons/Files.java
@@ -18,6 +18,7 @@ import java.nio.file.FileVisitResult;
 import java.nio.file.Path;
 import java.nio.file.SimpleFileVisitor;
 import java.nio.file.attribute.BasicFileAttributes;
+import java.util.List;
 import java.util.Optional;
 import java.util.regex.Pattern;
 
@@ -27,6 +28,18 @@ import java.util.regex.Pattern;
 public class Files {
 
     private static final Pattern NON_PATH_CHARACTERS = Pattern.compile("[^a-zA-Z0-9\\-.]");
+
+    /**
+     * Contains a list of file names and endings which are considered to be metadata.
+     */
+    private static final List<String> METADATA_FILES = List.of(
+            "__MACOSX",
+            "thumbs.db",
+            ".ini",
+            "$RECYCLE.BIN",
+            ".sys",
+            ".tmp",
+            ".temp");
 
     private Files() {
     }
--- a/src/main/java/sirius/kernel/commons/Files.java
+++ b/src/main/java/sirius/kernel/commons/Files.java
@@ -160,6 +160,34 @@ public class Files {
     }
 
     /**
+     * Determines if the given path is hidden.
+     * <p>
+     * A path is considered hidden if it starts with a dot.
+     *
+     * @param path the path to check
+     * @return <tt>true</tt> if the path is hidden, <tt>false</tt> otherwise
+     */
+    public static boolean isHidden(@Nullable String path) {
+        return path != null && path.startsWith(".");
+    }
+
+    /**
+     * Determines if the given path is a metadata file.
+     * <p>
+     * A metadata file is a file which is not part of the actual content but rather contains metadata or is used by
+     * the operating system or other tools.
+     *
+     * @param path the path to check
+     * @return <tt>true</tt> if the path is a metadata file that is listed in the METADATA_FILES list, <tt>false</tt> otherwise
+     */
+    public static boolean isMetadata(@Nullable String path) {
+        if (Strings.isEmpty(path)) {
+            return false;
+        }
+        return METADATA_FILES.stream().anyMatch(path::contains);
+    }
+
+    /**
      * If the given file is not null and exists, tries to delete that file and logs when a file cannot be deleted. T
      * his is useful for error reporting and to diagnose why a file cannot be deleted.
      *


src/main/java/sirius/kernel/commons/Files.java
```diff
@@ -28,6 +29,18 @@ public class Files {
 
     private static final Pattern NON_PATH_CHARACTERS = Pattern.compile("[^a-zA-Z0-9\\-.]");
 
+    /**
+     * Contains a list of file names and endings which are considered to be metadata.
+     */
+    private static final List<String> METADATA_FILES = List.of(
```
sabieber, line 35: Should these be Patterns like `.*\.ini`, ... to check more specifically? Some of them are prefixes, some are suffixes, some are entire file names.

src/main/java/sirius/kernel/commons/Files.java
```diff
@@ -146,6 +159,34 @@ public static String getFilenameWithoutExtension(@Nullable String path) {
...
 
+    /**
+     * Determines if the given path is hidden.
+     * <p>
+     * A path is considered hidden if it starts with a dot.
+     *
+     * @param path the path to check
+     * @return <tt>true</tt> if the path is hidden, <tt>false</tt> otherwise
+     */
+    public static boolean isHidden(@Nullable String path) {
```
idlira, line 170: **Suggestion**: rename to `isConsideredHidden`, since we are validating a string and not using any file-system feature.

src/main/java/sirius/kernel/commons/Files.java
```diff
@@ -146,6 +159,34 @@ public static String getFilenameWithoutExtension(@Nullable String path) {
...
+    /**
+     * Determines if the given path is a metadata file.
+     * <p>
+     * A metadata file is a file which is not part of the actual content but rather contains metadata or is used by
+     * the operating system or other tools.
+     *
+     * @param path the path to check
+     * @return <tt>true</tt> if the path is a metadata file that is listed in the METADATA_FILES list, <tt>false</tt> otherwise
+     */
+    public static boolean isMetadata(@Nullable String path) {
```
idlira, line 183: **Suggestion**: rename to isConsideredMetadata, since we are validating a string and not using any file-system feature.

sabieber (approved):

src/main/java/sirius/kernel/commons/Files.java
```diff
@@ -146,6 +159,34 @@ public static String getFilenameWithoutExtension(@Nullable String path) {
...
+        if (Strings.isEmpty(path)) {
+            return false;
+        }
+        return METADATA_FILES.stream().anyMatch(path::contains);
```
idlira, line 187: **Required**: sure we want to use contains against the list? This would discard real files with an extension `.ini` which are not necessarily metadata.
I think would be useful to get a hand of a few Windows-generated ZIPS and look what is inside. For macOS is clear and we can check with `.startsWith`

idlira (changes requested):

src/main/java/sirius/kernel/commons/Files.java
idlira: Or seeing Sacha's comment, maybe we need a RegEx match to solve that properly, or have at least 2 pattern lists, to check with startsWith and endsWith. It all depends on how these actually appear in the files.

src/main/java/sirius/kernel/commons/Files.java
```diff
@@ -146,6 +159,34 @@ public static String getFilenameWithoutExtension(@Nullable String path) {
...
+    /**
+     * Determines if the given path is hidden.
+     * <p>
+     * A path is considered hidden if it starts with a dot.
+     *
+     * @param path the path to check
+     * @return <tt>true</tt> if the path is hidden, <tt>false</tt> otherwise
+     */
+    public static boolean isHidden(@Nullable String path) {
+        return path != null && path.startsWith(".");
```
idlira, line 171: **Required:** One more thing, we are talking about a path here. Eg: `folder1/folder2/folder3/.hiddenfile`
We need to check if the final portion of the path is considered a hidden file. You can use Files.getFilenameAndExtension to obtain the last portion of it, and there check if it starts with `.`

idlira (changes requested):
One more note regarding validating longer path.
Perhaps a test case can be added to check these methods.

src/main/java/sirius/kernel/commons/Files.java
idlira: Yes, if more than one pattern exist, we can directly go with a list of patterns, which gives us way more flexibility.

src/main/java/sirius/kernel/commons/Files.java
```diff
@@ -28,6 +29,14 @@ public class Files {
 
     private static final Pattern NON_PATH_CHARACTERS = Pattern.compile("[^a-zA-Z0-9\\-.]");
 
+    /**
+     * Contains a list of file names and endings which are considered to be metadata.
+     */
+    private static final List<String> METADATA_FILES = List.of(
+            "__MACOSX",
+            ".DS_Store",
+            "thumbs.db");
```
idlira, line 38: I think Windows calls it `Thumbs.db`, (uppercase T). At least is what I find in some sample in our Google Drive storage.

src/main/java/sirius/kernel/commons/Files.java
```diff
@@ -146,6 +155,38 @@ public static String getFilenameWithoutExtension(@Nullable String path) {
...
+     * the operating system or other tools.
+     *
+     * @param path the path to check
+     * @return <tt>true</tt> if the path is a metadata file that is listed in the METADATA_FILES list, <tt>false</tt> otherwise
+     */
+    public static boolean isConsideredMetadata(@Nullable String path) {
+        if (Strings.isEmpty(path)) {
+            return false;
+        }
+        return METADATA_FILES.stream().anyMatch(path::startsWith);
```
idlira, line 187: Required: also apply here the same logic to obtain the last portion of the path.
The metadata folder or file could be in a lower level, like `folder1/folder2/__MACOSX`

idlira (changes requested):

src/main/java/sirius/kernel/commons/Files.java
```diff
@@ -147,8 +154,40 @@ public static String getFilenameWithoutExtension(@Nullable String path) {
...
+     * the operating system or other tools.
+     *
+     * @param path the path to check
+     * @return <tt>true</tt> if the path is a metadata file that is listed in the METADATA_FILES list, <tt>false</tt> otherwise
+     */
+    public static boolean isConsideredMetadata(@Nullable String path) {
+        if (Strings.isEmpty(path)) {
+            return false;
+        }
+        return streamPath(path).anyMatch(METADATA_FILES::contains);
```
mkeckmkeck, line 185: I would prefer working on path segments (files/folders) with equals comparision.

src/main/java/sirius/kernel/commons/Files.java
```diff
@@ -213,4 +251,26 @@ public FileVisitResult postVisitDirectory(Path dir, IOException exc) throws IOEx
...
+    /**
+     * Returns a stream of all parts of the given path.
+     * <p>
+     * The stream will contain the filename, the parent directory, the parent of the parent and so on.
+     *
+     * @param path the path to split
+     * @return a stream of all parts of the given path
+     */
+    public static Stream<String> streamPath(String path) {
+        if(Strings.isEmpty(path)) {
```
idlira, line 264: Code formatting?

idlira (approved):

src/main/java/sirius/kernel/commons/Files.java
```diff
@@ -147,8 +154,39 @@ public static String getFilenameWithoutExtension(@Nullable String path) {
...
+     * the operating system or other tools.
+     *
+     * @param path the path to check
+     * @return <tt>true</tt> if the path is a metadata file that is listed in the METADATA_FILES list, <tt>false</tt> otherwise
+     */
+    public static boolean isConsideredMetadata(@Nullable String path) {
+        if (Strings.isEmpty(path)) {
+            return false;
+        }
+        return streamPath(path).anyMatch(METADATA_FILES::contains);
```
idlira, line 184: This should now go back to .startsWith, since it only expects a target file/dir right?
PS: test cases must be than adjusted.

sabieber (approved):

PR #551, opened 2025-03-03
Adds a simple parallel task executor

idlira:
### Description

This is an easy to initialize and use executor which permits to run a number of limited tasks in parallel using virtual threads.
Useful for parallel tasks not demanding a complex clustered setup. For that, our framework already offers many options.

Call sample:
```java
ParallelTaskExecutor executor = new ParallelTaskExecutor(4);

for (int i = 0; i < 1000; i++) {
    final int taskId = i;
    executor.submitTask(() -> {
        System.out.println("Executing Task " + taskId + " on thread " + Thread.currentThread().getName());
        // Enough theory, do some real stuff in here...
    });
}

executor.shutdownWhenDone();
```

- tasks start running as soon as they are submitted
- shutdown will wait until the last task finishes
- current context is passed to each task

### Additional Notes

- This PR fixes or works on following ticket(s): [OX-11751](https://scireum.myjetbrains.com/youtrack/issue/OX-11751)

### Checklist

- [ ] Patch Tasks: Is local execution of Patch Tasks necessary? If so, please also mark the PR with the tag.

### DIFF
--- a/src/main/java/sirius/kernel/async/ParallelTaskExecutor.java
+++ b/src/main/java/sirius/kernel/async/ParallelTaskExecutor.java
@@ -1,0 +1,98 @@
+/*
+ * Made with all the love in the world
+ * by scireum in Remshalden, Germany
+ *
+ * Copyright by scireum GmbH
+ * http://www.scireum.de - info@scireum.de
+ */
+
+package sirius.kernel.async;
+
+import sirius.kernel.commons.Wait;
+
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.Semaphore;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Executes tasks in parallel in the current node using virtual threads with a limit on concurrency.
+ * <p>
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/kernel/async/ParallelTaskExecutor.java
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * Creates a new parallel task executor.
+     *
+     * @param maxConcurrentTasks the maximum number of tasks to run concurrently
+     */
+    public ParallelTaskExecutor(int maxConcurrentTasks) {
+        this.currentContext = CallContext.getCurrent();
+        this.executor = Executors.newVirtualThreadPerTaskExecutor();
+        this.taskQueue = new LinkedBlockingQueue<>();
```
mko-sci, line 46: Is it okay that the queue can still grow indefinitely since `LinkedBlockingQueue` is unbounded.
```
this.taskQueue = new LinkedBlockingQueue<>(maxConcurrentTasks * 2); // Limit queue size?
```

idlira: For the expected usage yes. If the need arises, we can add a different constructor to cope with a maximum queue size. For now I can add a hint to the JavaDoc explaining that the queue is only limited by your own RAM.

sabieber (approved):

jakobvogel (approved):

PR #592, opened 2026-01-16
Fixes HTML to plain text conversion

idlira:
### Description

`StringCleanup#htmlToPlainText` is expected to convert `p` and `br` tags to line breaks.
This works if these tags were not followed by other tags. For example:

➡️ Input: `Hello<br />World`
✅ Output:
```
Hello
World
```
but
➡️ Input: `<strong>Hello</strong><br />World`
🚫 Output: `HelloWorld`

This happens because `StringCleanup.PATTERN_STRIP_XML` will look for white-spaces leading or trailing a tag, effectively replacing `</strong>\n` in the example above by an empty string.

To not disrupt this pattern, and possible cause a breaking change/behavioral change, which is used in several other places, we iterate and clean each line.

### Additional Notes

- This PR fixes or works on following ticket(s): [OX-12298](https://scireum.myjetbrains.com/youtrack/issue/OX-12298)

### Checklist

### DIFF
--- a/src/main/java/sirius/kernel/commons/Strings.java
+++ b/src/main/java/sirius/kernel/commons/Strings.java
@@ -306,10 +306,10 @@ public class Strings {
     }
 
     /**
-     * Returns an url encoded representation of the given <tt>value</tt> with <tt>UTF-8</tt> as character encoding.
+     * Returns a url encoded representation of the given <tt>value</tt> with <tt>UTF-8</tt> as character encoding.
      *
      * @param value the value to be encoded.
-     * @return an url encoded representation of value, using UTF-8 as character encoding.
+     * @return a url encoded representation of value, using UTF-8 as character encoding.
      * @deprecated use {@link Urls#encode(String)} instead.
      */
     @Nullable
@@ -319,10 +319,10 @@ public class Strings {
     }
 
     /**
-     * Returns an url decoded representation of the given <tt>value</tt> with <tt>UTF-8</tt> as character encoding.
+     * Returns a url decoded representation of the given <tt>value</tt> with <tt>UTF-8</tt> as character encoding.
      *
      * @param value the value to be decoded.
-     * @return an url decoded representation of value, using UTF-8 as character encoding.
+     * @return a url decoded representation of value, using UTF-8 as character encoding.
      * @deprecated use {@link Urls#decode(String)} instead.
      */
     @Nullable
@@ -499,7 +499,7 @@ public class Strings {
     /**
      * Returns a string representation of the given map.
      * <p>
-     * Keys and values are separated by a colon (:) and entries by a new line.
+     * Keys and values are separated by a colon {@code :} and entries by a new line.
      *
      * @param source to map to be converted to a string
      * @return a string representation of the given map, or "" if the map was null
@@ -633,7 +633,7 @@ public class Strings {
      * <p>
      * Note that empty/<tt>null</tt> inputs will always result in an empty string.
      *
-     * @param inputString the string to clean-up
+     * @param inputString the string to clean up
      * @param cleanups    the operations to perform, most probably some from {@link StringCleanup}
      * @return the cleaned up string
      * @see StringCleanup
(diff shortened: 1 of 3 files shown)


src/main/java/sirius/kernel/commons/Strings.java
```diff
@@ -861,4 +863,21 @@ private static String performPadding(@Nullable String input, @Nonnull String pad
...
+     * @param lineConsumer the consumer to be applied for each line. The first parameter is the line number (starting at 1),
+     */
+    public static void iterateLines(String text, BiConsumer<Integer, String> lineConsumer) {
+        try (Scanner scanner = new Scanner(text)) {
+            int lineNumber = 0;
+            while (scanner.hasNextLine()) {
+                lineConsumer.accept(++lineNumber, scanner.nextLine());
+            }
+        }
+    }
```
jakobvogel: There is [String.lines()](https://docs.oracle.com/en/java/javase/24/docs/api/java.base/java/lang/String.html#lines()) in the standard library.

jakobvogel (approved):

mkeckmkeck (approved):

src/main/java/sirius/kernel/commons/Strings.java
idlira: Cool. Have replaced but will keep the method which also deliver the line number, so I can replace in the original place I actually cloned this method from.

idlira: ... but can drop it at all if you find it overkill or not much functionality added.

jakobvogel: I am fine if you want to keep the method. I personally would remove it, as I don't see much use for the line number.

In your case, you could potentially also check for emptiness, i.e. in pseudo-code

```
if (!builder.isEmpty()) {
    builder.append("\n");
}
```

However, and this may be good to define via unit tests, we should consider special cases such as

```
<br />
<br />
Hello<br />
World
```

jakobvogel: (Saw your latest commit only now, my review was interrupted by a support case.)

ymo-sci (approved):

src/test/kotlin/sirius/kernel/commons/StringsTest.kt
```diff
@@ -425,5 +425,12 @@ class StringsTest {
             )
         )
 
+        assertEquals(
+            "bold\nmove",
+            Strings.cleanup(
+                "<strong>bold</strong><br />move",
+                StringCleanup::htmlToPlainText
+            )
+        )
```
jakobvogel, line 434: As mentioned in my other comment: Maybe also add special cases as tests, such as the following one.

```
<br />
<br />
Hello<br />
World
```

And I know, the test method was already like that before, but IMHO this is a case for a parameterized test. Right now, the list of unrelated tests in one method gives less reliable statistics if one of the tests fails.

jakobvogel (approved):

src/test/kotlin/sirius/kernel/commons/StringsTest.kt
idlira: These exist already: 

```
        assertEquals(
            """
            first

            second
        """.trimIndent(),
            Strings.cleanup("<p>first<br><br/>second</p>", StringCleanup::htmlToPlainText, StringCleanup::trim)
        )
```

jakobvogel: That test is not equivalent to my example, as it does not contain line breaks, but it includes trimming. Therefore, it does not test the behaviour of your new approach. 🙃

I played around with this a little:

- An extension of your test case by adding a line break: `<strong>bold</strong><br />\nmove` gives `bold\n move` (with a space in front of `move`)
- My example `<br />\n<br />\nHello<br />\nWorld` gives ` \n Hello\n World` (with a space at the front of each of the **three** lines!)

I am not sure whether these really are the results we want. Therefore, I would appreciate to have respective additional test cases defining the gold standard.

idlira: the logic in this class is so broken... will need review and posterior PR

PR #593, opened 2026-01-16
Fixes HTML to plain text conversion... again

idlira:
### Description

**Note in advance:**
we should evaluate using **jsoup** or some lib to perform these kind of operations. But since that required an analysis (impact and feasibility), I just try to improve the current code. There is no real recognition of things like inline vs block elements.

The HTML code might contain line-breaks. These are not considered for display and the actual `br` and `p` tags must be use to really display a line-break.

In addition, HTML reduces spaces. A string like `"<b>Hello  </b>   world"` is displayed as "**Hello** world".

### Additional Notes

- This PR fixes or works on following ticket(s): [OX-12298](https://scireum.myjetbrains.com/youtrack/issue/OX-12298)
- This PR is related to PR: https://github.com/scireum/sirius-kernel/pull/592

### Checklist

### DIFF
--- a/src/main/java/sirius/kernel/commons/StringCleanup.java
+++ b/src/main/java/sirius/kernel/commons/StringCleanup.java
@@ -19,6 +19,7 @@ import java.util.Map;
 import java.util.TreeMap;
 import java.util.function.UnaryOperator;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 /**
  * Provides various methods to clean-up or reduce strings.
@@ -181,7 +182,7 @@ public class StringCleanup {
                                                                       TAG_UL,
                                                                       TAG_LI);
 
-    private static final Pattern PATTERN_STRIP_XML = Pattern.compile("\\s*" + Strings.REGEX_DETECT_XML_TAGS + "\\s*");
+    private static final Pattern PATTERN_STRIP_XML = Pattern.compile(Strings.REGEX_DETECT_XML_TAGS);
     private static final Map<Integer, String> unicodeMapping = new TreeMap<>();
 
     static {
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/kernel/commons/StringCleanup.java
```diff
@@ -546,11 +547,11 @@ public static String decodeHtmlEntities(@Nonnull String input) {
      */
     @Nonnull
     public static String htmlToPlainText(@Nonnull String input) {
-        String normalizedText = input;
+        if (PATTERN_STRIP_XML.matcher(input).find()) {
+            // Start joining all lines. A browser does not recognize line breaks in the HTML source as these
+            // are just whitespaces. Handling of actual HTML line breaks happens in the following lines.
+            String normalizedText = input.lines().collect(Collectors.joining());
```
fhaScireum: Do browsers actually ignore line breaks or do they treat them as an empty space? If the latter, wouldn't we need `Collectors.joining(" ")` here?

idlira: Ignore
``` html
<html>
<body>
<br />
<br />




Hello<br />


World
<p>
    hello
</p>
<b>Hello  </b>   world
</body>
</html>
```
image

jakobvogel (approved):
Thank you, @idlira!

I would still prefer to do that all via jsoup, particularly the conversion of `<p>` seems to be strange judging from the tests that require trimming. Also, I would still prefer parameterized tests. But I agree: For now, this is fine.

src/main/java/sirius/kernel/commons/StringCleanup.java
sabieber: Your sample does not include multiple inline elements on multiple lines, like:

```
<span>Line 1</span>
<span>Line 2</span>

**Review by sabieber — APPROVED** [review]

**Review by mkeckmkeck — APPROVED** [review]

**Review by sabieber — APPROVED** [review]

File: src/test/kotlin/sirius/kernel/commons/StringsTest.kt [hunk 2]
```diff
@@ -445,5 +445,21 @@ class StringsTest {
...
+            Strings.cleanup(
+                "<span>Span 1</span>\n<span>Span 2</span>",
+                StringCleanup::htmlToPlainText
+            )
+        )
+
+        assertEquals(
+            "\nLine 1 Line 2 Line 3",
+            Strings.cleanup(
+                "<p>Line 1\nLine 2\nLine 3</p>",
```
jakobvogel, line 460: I know, it seems to be legacy behaviour, and we don't need to change this now: But for future extensions, we should consider whether `<p>…</p>` should really cause a leading newline character in the plain text version. IMHO, this should only be there if there is other content before the `<p>`

PR #634, opened 2026-08-04
Keeps a StructuredOutput's result as a JsonNode instead of writing it out

jakobvogel:
### Description

Adds `JsonNodeStructuredOutput`, an implementation of `StructuredOutput` which keeps its result in memory as a
`JsonNode` instead of writing it out.

Every implementation so far writes the result somewhere as it is produced — `XMLStructuredOutput` into a
document, sirius-web's `JSONStructuredOutput` into a response body. So code written to describe a payload can
only be used where that payload *is* the answer. Wanting the same payload as a **value** — to embed it into a
larger document, to hand it to a client library, to compare it against an expectation in a test, or to
post-process it before sending — meant describing it a second time, and two descriptions of one thing drift
apart without anybody noticing.

```java
ObjectNode node = JsonNodeStructuredOutput.collect(output -> {
    output.property("id", customer.getId());
    output.property("name", customer.getName());
});
```

Values render as the streaming JSON output renders them, so the two agree on dates, enums, amounts and what
stays unquoted.

**Amounts are the one place where they deliberately differ.** `JSONStructuredOutput` writes the formatted value
straight into the document, so an amount formatted with a localized `NumberFormat` ("1.234,50") produces
invalid JSON there. Here a machine-formatted amount becomes a number and anything else is kept verbatim as a
string, so no caller can produce a document a parser would refuse. Both directions are covered by a test.

No existing behaviour is touched: this only adds a class, its test and a paragraph in the package README.

### On the package, which is worth a word

The class lands in `sirius.kernel.xml`, and a JSON implementation in a package named `xml` reads oddly. That is
deliberate rather than careless, and I would rather point at it than let it look accidental:

- The package is where the format-neutral part of the abstraction already lives — `StructuredOutput`,
  `StructuredInput`, `AbstractStructuredOutput`, `Attribute` — so the implementation sits with the contract it
  implements.
- `sirius.kernel.commons`, next to `Json` and `CSVWriter`, would read better. But the class extends
  `AbstractStructuredOutput`, so putting it there would introduce the first `commons` → `xml` package
  dependency; today the arrow points only the other way.
- A neutral package would be the honest fix, but only if `StructuredOutput` and friends moved into it as well,
  and that breaks every import in every downstream product.

For the same reason it is worth noting where the sibling implementation is: `JSONStructuredOutput` lives in
`sirius.web.services`, although it depends on nothing from sirius-web — only on `sirius.kernel.*`, `java.io` and
Jackson. So one interface currently has implementations in three places, and this PR makes that three rather
than two.

If you would rather have it elsewhere, say where and I will move it — the naming is yours to decide, and none
of this is load-bearing for the change itself.

### Additional Notes

- No ticket — this came out of a downstream product that needed the same payload description for a response and
  for a value, and the gap is general enough to belong here rather than in the product.

### Checklist

- [ ] Code was formatted via IntelliJ and follows SonarLint & [best practices](https://scireum.myjetbrains.com/youtrack/articles/MISC-A-16/CodeStyle-JavaDoc)
- [x] Patch Tasks: Is local execution of Patch Tasks necessary? If so, please also mark the PR with the tag.

### DIFF
--- a/src/main/java/sirius/kernel/xml/JsonNodeStructuredOutput.java
+++ b/src/main/java/sirius/kernel/xml/JsonNodeStructuredOutput.java
@@ -1,0 +1,206 @@
+/*
+ * Made with all the love in the world
+ * by scireum in Remshalden, Germany
+ *
+ * Copyright by scireum GmbH
+ * http://www.scireum.de - info@scireum.de
+ */
+
+package sirius.kernel.xml;
+
+import sirius.kernel.commons.Amount;
+import sirius.kernel.commons.Json;
+import sirius.kernel.commons.NumberFormat;
+import sirius.kernel.commons.Strings;
+import tools.jackson.databind.JsonNode;
+import tools.jackson.databind.node.ArrayNode;
+import tools.jackson.databind.node.JsonNodeFactory;
+import tools.jackson.databind.node.ObjectNode;
+
+import java.math.BigDecimal;
+import java.util.ArrayDeque;
+import java.util.Deque;
+import java.util.function.Consumer;
+
+/**
+ * Assembles a {@link JsonNode} in memory instead of writing to a stream.
+ * <p>
+ * The implementations of {@link StructuredOutput} write their result somewhere as it is produced: into an XML document
+ * ({@link XMLStructuredOutput}) or into a response body (<tt>JSONStructuredOutput</tt> in <tt>sirius-web</tt>). This
+ * one keeps it, so that code written to describe a payload can also be used where the payload is needed as a
+ * <b>value</b> — embedded into a larger document, handed to a client library, compared against an expected result in a
+ * test, or post-processed before it is sent.
+ * <p>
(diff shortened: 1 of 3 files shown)


sabieber (approved):
I don't fully get why we need this, yet, as one can directly use a `JsonGenerator`. But it doesn't hurt either.

fhaScireum (approved):

idlira (approved):

jmuscireum (approved):

jakobvogel:
> I don't fully get why we need this, yet, as one can directly use a `JsonGenerator`. But it doesn't hurt either.

We have many helpers and API routes that currently take a `JSONStructuredOutput`. The easiest way to re-use that without major refactoring is this adapter. it is formally a structured output (hence a valid parameter for the legacy methods), and it provides an accessible JSON node that I can use elsewhere afterward, like within an MCP's RPC stuff.

